Issue #5, opened 2018-04-15
docker-compose up causes "manifest for nicolasdorier/btcpayserver:1.0.1.83 not found"


bitmario:
```
root@btcpayvm:~/btcpayserver-docker# docker-compose -f "$(pwd)/Production/docker-compose.btc-clightning.yml" up -d
[...]
Pulling btcpayserver (nicolasdorier/btcpayserver:1.0.1.83)...
ERROR: manifest for nicolasdorier/btcpayserver:1.0.1.83 not found
```

This tag doesn't seem to exist, so I fixed it by changing btcpayserver in the yml and setting:
`image: nicolasdorier/btcpayserver:1.0.1.84`

PR #6, opened 2018-04-15
Update btcpayserver version in yml

bitmario:
As described in #5 the btcpayserver 1.0.1.83 tag does not exist which causes docker-compose up to fail, so I'm updating to latest version (1.0.1.84)


NicolasDorier:
yes sorry I messed up :(

PR #11, opened 2018-06-14
Feathercoin Support

ChekaZ:


NicolasDorier:
Use tabs and use `git update-index --chmod=+x feathercoin-cli.sh` and on the ps1.

PR #13, opened 2018-06-18
Add FTC to docker-compose-generator

lclc:
If you could re-run the docker-compose-generator it would add FTC as one of the options for btcpay-setup.sh (like BTG is), as far as I understand the setup.

Thanks for your work. We plan to use BTCPay quite a bit to push merchant adoption for BTC, LTC & FTC in Switzerland and to contribute back whenever possible :)


NicolasDorier:
The `launchsettings` are just here for debugging. When you run the project with F5 in Visual studio, then it would try with FTC enabled.

PR #16, opened 2018-07-22
Add btcpay-down tool

hubiktomas:
For purposes of some monitoring tools that require both start and stop scripts.


NicolasDorier:
Merging, next time think about running `git update-index --chmod=+x btcpay-down.sh` so that the file is executable when someone git pull.

hubiktomas:
Thanks Nicolas, sorry for that. Will do next time. :)

PR #30, opened 2018-09-14
Configurable no reverse proxy port

### LINKED ISSUE #29: Specify nginx web ports to allow apache/other reverse proxying
I want to use btcpay behind my existing apache installation. Any way to specify alternatives for nginx other than 80,443?

Kukks:
Allows you to specify a BTCPAY_PORT var before you run btcpay-setup.sh in case you'd like to change the exposed port when not using a reverse proxy Closes #29


Kukks:
I've just introduced a new optional env variables `BTCPAY_INTERNAL_PORT` which sets which port the btcpayserver binds the app to. This is then also configured all the way through to nginx

NicolasDorier:
Why do we need `BTCPAY_INTERNAL_PORT` ?

btcpay-setup.sh
```diff
@@ -196,6 +198,11 @@ if [ -z "$BTCPAY_HOST" ]; then
     return
 fi
 
+if [ -z "$BTCPAY_PORT" ]; then
+    echo "BTCPAY_PORT defaulting to 80"
+	$BTCPAY_PORT = 80
```
NicolasDorier: indent

Kukks:
What usecase do we need BTCPAY_INTERNAL_PORT ?

> With some reverse proxies when used in docker in conjunction with specific network settings, they hook up with the actual port inside the container. At that point i'ts good to be able to organize which used ports you have

btcpay-setup.sh
```diff
@@ -196,6 +198,11 @@ if [ -z "$BTCPAY_HOST" ]; then
     return
 fi
 
+if [ -z "$BTCPAY_PORT" ]; then
+    echo "BTCPAY_PORT defaulting to 80"
+    $BTCPAY_PORT = 80
+fi
+
```
NicolasDorier: You can replace all the block with `: "${BTCPAY_PORT:=80}"`

NicolasDorier:
I still don't understand why you need `BTCPAY_INTERNAL_PORT`. What is the big deal with hard coding it? Less flexibility means less headache and loose screw.

btcpay-setup.sh
```diff
@@ -49,6 +49,7 @@ This will be used to properly setup HTTPS via let's encrypt.
 
 Environment variables:
     BTCPAY_HOST: The hostname of your website (eg. btcpay.example.com)
+    BTCPAY_PORT: The port to expose (80 by default)
```
NicolasDorier: If not using any reverse proxy, the port to expose (80 by default)

btcpay-setup.sh
```diff
@@ -163,6 +164,7 @@ echo "
 -------SETUP-----------
 Parameters passed:
 BTCPAY_HOST:$BTCPAY_HOST
+BTCPAY_PORT:${BTCPAY_PORT:=80}
```
NicolasDorier: I am wondering if we can have a one liner to not put this line if BTCPAY_PORT is null or unset. It might be confusing for a user using nginx to see `BTCPAY_PORT: 80` when his website is using 443.

NicolasDorier:
No, what you just did do not work: : just replace the three lines with `: "${BTCPAY_PORT:=80}"`, keeping the rest the same.

Kukks:
I'll get to this one later with a fresh branch

PR #36, opened 2018-09-21
Bump Groestlcoin Core to v2.16.3

gruve-p:
New Groestlcoin Core with CVE-2018-17144 patch


NicolasDorier:
I bumped it myself, can you try to update?

PR #41, opened 2018-09-25
add in lnd auto pilot for all other lnd supporting coins and add in pruning options for all other coins

### LINKED ISSUE #40: Expand pruning to other crypto nodes
We need to allow pruning to other crypto nodes. Guessing we should add a key to the crypto definition file at this point? We should also duplicate the exisitng btc ones to have fragments starting with `btc-* ` to scale. :)

Kukks:
closes #40

### DIFF
--- a/docker-compose-generator/docker-fragments/opt-bgold-lnd-autopilot.yml
+++ b/docker-compose-generator/docker-fragments/opt-bgold-lnd-autopilot.yml
@@ -1,9 +1,0 @@
-version: "3"
-
-services:
-  lnd_bitcoingold:
-    environment:
-      LND_EXTRA_ARGS: |
-        autopilot.active=1
-        autopilot.maxchannels=5
-        autopilot.allocation=0.6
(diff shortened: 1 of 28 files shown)


NicolasDorier:
@Kukks don't create separate fragment per coins. Just add stuff to `opt-lnd-autopilot`. If the image is not present, it will just be ignored.

Kukks:
Are you sure that works? The last time I tried that it caused issues for when you didn't include the service. For example, having `opt-lnd-autopilot` handle both btc and ltc when you did not include ltc would cause the compose file to be invalid

NicolasDorier:
@Kukks I might have fixed that since then... else we need to just ignore so it does not crash.

Kukks:
Wouldn't having them separated be good in any case though? It allows users to be able to decide what level of pruning they want to achieve per coin along with the type of lightning integration?

NicolasDorier:
@Kukks if the user wants its own rules, he can use custom fragments.
I think you should have only 1. If they want more complex rule per crypto, it will be very easy for them to copy part of what they want and make their own fragments.

NicolasDorier:
Different crypto have different block time and size, so in theory we should adjust the numbers... but well, since most altcoins are unused it won't matter too much I think.

PR #44, opened 2018-10-01
Updating runtime image, required for BtcPayServer bugfix

rockstardev:
Moving to ASPNET Core 2.1.1


NicolasDorier:
Yep I think it is best.

NicolasDorier:
Closing this, needed to update the SDK as well as the AspNet meta package.

I also updated all images NBXplorer and BTCPayServer.

PR #58, opened 2018-11-13
Better Fragment  support

Kukks:
This allows us to have custom fragments that update the image used on the main node. It also allows us to overwrite the main node's scalar values through custom fragments when needed.

### DIFF
--- a/docker-compose-generator/src/DockerComposeDefinition.cs
+++ b/docker-compose-generator/src/DockerComposeDefinition.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Linq;
 using System.Collections.Generic;
 using System.Text;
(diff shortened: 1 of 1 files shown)


NicolasDorier:
If there is two yml files with an `image` field, which one is considered the main node?

NicolasDorier:
If there is two yml files with an `image` field, which one is considered the main node?

Kukks:
It grabs the first one in the order

NicolasDorier:
so which one? The order is not deterministic.

Kukks:
Order is hardcoded to the following(current untouched code):
* proxy fragments
* btcpayserver
* nbxplorer
* postgres
* crypto definitions
* additional fragments in order

NicolasDorier:
I don't like this, it depends on the order, this is quite dangerous I think, as I make no guarantee on this.
Now we have the `BTCPAY_IMAGE` you can change the image.

Kukks:
It is also guaranteed from the hard coded order though? additional fragments always run last like this

I find this one useful for more than just versioning management though. This allows you to overwrite scalar values which could come in handy across some of the services.

NicolasDorier:
Overriding the scalars is one thing. But having a second node with the image node does not seem right.

Also overriding the scalar is strange, as in the case of mappings, I concatenate the values instead of overriding.

Have you a real use case of this?

Kukks:
No other usecases for now, I am using the new btcpay image flag anyways. Just seemed like it could be useful in case a user wanted to replace some default hardcoded settings without forking

PR #71, opened 2018-12-03
Add Dash 12.3 Support

snogcel:
Adds support for Dash 0.12.3 on Linux/AMD64.

Corresponding Docker Image: https://github.com/btcpayserver/dockerfile-deps/pull/1


NicolasDorier:
Use `btcpayserver/dash:0.12.3` so that it works on all platforms.

snogcel:
updated!

PR #72, opened 2018-12-04
Isolate postgres and enable mysql & sqlite through docker

Kukks:
This PR isolates the logic related to postgres completely to its own fragment and adds a new fragment to use a mysql db if chosen.


docker-compose-generator/docker-fragments/mysql.yml
```diff
@@ -0,0 +1,19 @@
...
+    restart: unless-stopped
+    image:  mysql:8.0
+    volumes:
+      - "mysql_datadir:/var/lib/mysql"
+    environment:
+      MYSQL_ALLOW_EMPTY_PASSWORD: yes
+
+  btcpayserver:
+    environment:
+      BTCPAY_POSTGRES: Server=mysql;Database=btcpayserver${NBITCOIN_NETWORK:-regtest};Uid=root;Pwd=;Port=3306;
```
NicolasDorier: BTCPAY_MYSQL I think

NicolasDorier:
Might interest @sipsorcery

sipsorcery:
Happy to test. Fee free to ping me when ready.

Kukks:
@sipsorcery  I have just deployed it to https://btcpay.kukks.org and it seems ok. If you wnat to test it before merge, you;ll need to build the `btcpayserver/docker-compose-generator` manually locally( there is an issue with the `btcpayserver/docker-compose-generator:local` flag #73)

@NicolasDorier

docker-compose-generator/src/Program.cs
```diff
@@ -51,7 +51,9 @@ private void Run(DockerComposition composition, string name, string output)
 			}
 			fragments.Add("btcpayserver");
 			fragments.Add("nbxplorer");
-			fragments.Add("postgres");
```
NicolasDorier: where is `postgres`?

docker-compose-generator/src/Program.cs
```diff
@@ -51,7 +51,9 @@ private void Run(DockerComposition composition, string name, string output)
...
+			if(composition.SelectedDatabase != "sqlite"){
```
NicolasDorier: format

NicolasDorier:
no WIP anymore?

Kukks:
no longer wip @NicolasDorier  Seems to work from a deployment i launched on btcpay.kukks.org

sipsorcery:
Is the idea that the docker compose file can be generated on Windows as well as Linux (I can't see the PowerShell equivalent to btcpay-setup.sh)? 

And on Windows how do the build.ps1 script and the docker-compose-generator VS project work together or are they two different ways to generate the custom docker-compose file?

I have just enough docker expertise to be dangerous and am happy to test things out but I may take a bit of teaching.

NicolasDorier:
`build.ps1` is pulling image from docker, not using your local stuff. I don't remember how to use the local one. I think it is documented in readme, but unsure if it still work.

NicolasDorier:
I needed to revert it. It broke everything during the update.

Excisting users were using sqlite suddenly.

NicolasDorier:
error during the build
```
        btcpayserver
        nbxplorer
         not found in /app/docker-fragments/.yml, ignoring...
        litecoin
```

docker-compose-generator/src/DockerComposition.cs
```diff
@@ -39,6 +45,7 @@ public static DockerComposition FromEnvironmentVariables()
 					continue;
 				composition.SelectedCryptos.Add(selectedCrypto.ToLowerInvariant());
 			}
+			composition.SelectedDatabase = (Environment.GetEnvironmentVariable("BTCPAYGEN_DATABASE") ?? "postgtres").ToLowerInvariant();
```
NicolasDorier: `postgtres` typo + if empty string, then it will fails not use `postgtres`

Kukks:
Meh I should have done an enumeration to avoid magic strings, sorry about
that

PR #75, opened 2018-12-07
Feature/mysql

Kukks:
# fixes the deadly typo

* BTCPAYGEN_DATABASE flag for `postgres`, `mysql` & `sqlite`
* Defaults to postgres
* Also added in the small fix to build the generator locally(set to amd64 only for now, not a lot of people modify this stuff on an arm device ) when local flag is provided


docker-compose-generator/src/DockerComposition.cs
```diff
@@ -39,6 +45,7 @@ public static DockerComposition FromEnvironmentVariables()
 					continue;
 				composition.SelectedCryptos.Add(selectedCrypto.ToLowerInvariant());
 			}
+			composition.SelectedDatabase = (Environment.GetEnvironmentVariable("BTCPAYGEN_DATABASE") ?? "postgres").ToLowerInvariant();
```
NicolasDorier: Does not cover the case if `BTCPAYGEN_DATABASE` exist but is empty string.

NicolasDorier:
please try everything on your own node with postgres. This is a tricky update.

Kukks:
I am trying to think of the best way to roll out this upgrade, it's a bit hard since we have changes in both existing fragments and the generator... 
Maybe we can push the generator and fragments first, build it in hub and then release the other tooling that enables database selection

Kukks:
I will cancel this PR. Offering Mysql here would just result in people expecting first-grade support for it.

PR #88, opened 2019-01-26
Quickbooks Update for New "taxIncluded" Field

JeffVandrewJr:
Now that BTCPay supports `taxIncluded` on invoices, BTCQBO is no longer limited to users who send invoices directly from Quickbooks.

Most online merchants do not invoice directly from Quickbooks, so before `taxIncluded` was integrated, the plugin had no use to them.

BTCQBO now has two "modes":

1. Invoice Mode: This was previously the only mode. This is for users who send invoices directly from Quickbooks.

2. Deposit Mode. This was enabled by `taxIncluded`. Now merchants that do not send invoices directly from Quickbooks can simply have Quickbooks record all BTCPay paid invoices directly into Quickbooks, with revenue and sales tax appropriately apportioned.

In Deposit Mode, merchants must set the Quickbooks Plugin as `notificationUrl`, however, they can set an IPN Forwarding URL in the plugin itself. After the plugin receives the IPN, it will forward it to its ultimate destination.

So a hypothetical online merchant would have the relevant IPN sent to BTCQBO where Quickbooks would record the payment, and BTCQBO would then forward the IPN to the relevant merchant solution so that the merchant knows that the relevant items could be shipped.

### DIFF
--- a/docker-compose-generator/docker-fragments/opt-add-btcqbo.yml
+++ b/docker-compose-generator/docker-fragments/opt-add-btcqbo.yml
@@ -2,7 +2,7 @@ version: "3"
 
 services:
   btcqbo:
-    image: jvandrew/btcqbo:0.1.20
+    image: jvandrew/btcqbo:0.2.4
     environment:
       REDIS_URL: "redis://redis:6379/0"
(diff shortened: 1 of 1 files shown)


NicolasDorier:
ah this beg a video! The deposit mode seems very powerful :))

PR #93, opened 2019-01-29
Version Update for BTC-QBO

JeffVandrewJr:
Version 0.3.24 of BTC-QBO contains backend efficiency improvements:

1. The RQ-Worker container is no longer necessary, as Quickbooks Oauth refresh is now accomplished via scheduled interval jobs stored directly to Redis.

2. Processing of IPNs is handled more efficiently.

These changes are designed to minimize the amount of system resources required by BTC-QBO. It should be as "light" as possible when doing its job.


NicolasDorier:
Super cool, I hated this additional container.

PR #101, opened 2019-02-21
bump upload size

Kukks:


NicolasDorier:
why?

Kukks:
plugins and theme files on woocommerce are almost always bigger than the default.

PR #109, opened 2019-03-04
bump dash docker image to 0.13.0

snogcel:


snogcel:
Just FYI, Dash Core v0.13.1 contains fixes related to mining and are not necessary for typical full node operation and is not included in this update. 

Dash Core v0.13.2 will be released in the near future which I plan to track in a subsequent update to dockerfile-deps as well as btcpayserver-docker.

Issue #116, opened 2019-03-21
Create automated backup


Kukks:
Leverage https://github.com/blacklabelops/volumerize

JeffVandrewJr:
This is super cool. Dropbox and Google Drive built in.

astupidmoose:
Yes! Backups are a must. 

I don't think it's important to backup the nodes as they can be resyned, but postgres at a minimum should be automatically backed up.

JeffVandrewJr:
Hey guys,

I've been messing around with various ideas for backups and have some thoughts.

The volumerize tool is cool, but since our users have a variety of different volumes depending on their setup (different lightning instances, plugins, etc). This would require a lot of changes to the `docker-compose-generator` to get right,. Volumizer also starts and stops containers, which is an issue with the `docker-compose` restart policies. [The volumizer docs mention explicitly that this will create conflicts.]

I think it's actually easier to do this with good old fashioned shell scripting. I spent this afternoon coding a very simple backup solution that has two pieces:

1. One new shell script in the `btcpay-docker` directory, called `backup.sh`. This script runs `btcpay-down.sh`, then creates a tarball of `/var/lib/docker` (excluding the bitcoind data dir since it's so huge). Then runs `btcpay-up.sh`. After `btcpay-up.sh`, it hands off the tarball to one of the cloud storage clients

2. A new subdirectory in `btcpay-docker` called `btcpay-backup` would contain clients for Dropbox, Google Drive, Amazon S3, etc. Right now for my testing I have already written the Dropbox client and it works smoothly. All it needs is a Dropbox access token which the user obtains from the Dropbox website. The client reads the access token from an environmental variable. 

The idea would be to have `backup.sh` run via a cron job in the middle of the night. On my testing node, the total downtime required was about 3 minutes. On a "real" instance with lots of transactions, this could of course be higher.

Obviously the tarball would need to be encrypted before sending "real" data to the cloud. I'm open for suggestions on encryption methods that balance security with ease to the user. If `gpg` is used, we somehow need to get the user to either keep his key safe or use the `-c` flag to encrypt via user-selected password (easier for the user, but they'll choose dumb passwords).

I don't have this stuff up on GitHub yet and I'm heading out to dinner now. If the consensus is that this is a decent approach, I'll refine it and make a PR for you all to review. 

[...]

PR #117, opened 2019-03-25
Add Automated Backup

JeffVandrewJr:
My description didn't carry, I'll add it below in a sec.

EDIT: See description and caveats below ⬇️


JeffVandrewJr:
This is a response to #116 @Kukks @NicolasDorier.

Adds the following:

`backup.sh`: Stops all containers cleanly, creates a tarball of `var/lib/docker` excluding the bitcoin blockchain, then hands off the tarball to one of many potential cloud uploader scripts.

`dropbox-script`: This is a cloud upload script for Dropbox that I threw together. The subdirectories are just all of the script dependencies; they are minimal but I included them directly in the repo so that they'll automatically come along when a user runs `btcpay-update.sh`.

Right now Dropbox is the only uploader script that's actually written, but `backup.sh` is written such that many others could follow.

The following environmental variables are expected to be set:

`BACKUP_PROVIDER`: Right now the only one that works is Dropbox.
`DROPBOX_TOKEN`: The Oauth token for Dropbox. The token can easily be obtained at developers.dropbox.com, Create App, give it access only to the special app folder, and then click the button on the dashboard to "generate token."

I presume ultimately you guys will set those env variables through the web interface, or you may bag my Dropbox client altogether for something else you're building.

Kukks:
whoa, that is a lot of new files

JeffVandrewJr:
@Kukks Those 100 files are the Dropbox SDK library and its dependencies. I included them whole since `btcpay-update` uses git on this repo to pull all the files it needs.

Without those, this is just two files: `backup.sh` and `dropbox-script`. But then you'd have to install the rest of the Dropbox dependencies through `pip`. Or I could compile them all together as a static binary.

If you're not using my Dropbox script than it wouldn't matter at all and you'd just need `backup.sh`.

NicolasDorier:
image
omfg.

NicolasDorier:
> Without those, this is just two files: backup.sh and dropbox-script. But then you'd have to install the rest of the Dropbox dependencies through pip

No, just need to use a python docker image.

NicolasDorier:
I like this.
Though the best would be that it was encapsulated inside a docker image.

The host would just need to run the docker image and bind the volumes folders.

NicolasDorier:
@JeffVandrewJr Can you dockerify it and put it to https://github.com/btcpayserver/dockerfile-deps ?

Don't need to do the ARM part if you don't have the time, I can do it once you have the amd64 working.

JeffVandrewJr:
@NicolasDorier Sure, I can do that.

In my head I didn't run it in Docker to begin with  because I was thinking Docker would be shut down for the backup, but I was being stupid because obviously I'm bringing Docker back up before the upload starts anyway.

NicolasDorier:
@JeffVandrewJr I am not even sure that it is worth it to shut down docker containers.

JeffVandrewJr:
@NicolasDorier 

Dockerizing was easy so I made the changes before bed. See the changes on this PR and also https://github.com/btcpayserver/dockerfile-deps/pull/4

My script stops the containers briefly because I think there is value in taking a static backup of the lightning folders. I believe I've had conversations with @PierreRochard about his Node Launcher where he takes backups of lightning folders upon shutdown for this reason. [Pierre if you see this and I misremember, correct me!]

Lightning aside, it seems like there would be less chance of getting a backup of corrupt data by stopping everything for a few minutes. But if I'm off base, we can just remove that line from `backup.sh` and we'd be good to go with no stoppage.

Changes made to file locations:
- I removed my Dropbox client from the btcpayserver-docker/repo
- I moved my Dropbox client to the dockerfile-deps repo

Changes made to `backup.sh`:
- The script now checks to see if a Docker volume exists for the Dropbox client. If it doesn't, it uses `docker volume create` to make one. 
- After that, it sends the tarball directly into the Docker volume
- It then launches the Dropbox container
- Post-upload, the Dropbox container shuts itself down gracefully

NicolasDorier:
https://github.com/btcpayserver/dockerfile-deps/pull/4


image

NicolasDorier:
@JeffVandrewJr I think the image for dropbox is perfect. However, you should allow the user to put his own file name in case he wants to keep multiple backups.

NicolasDorier:
Ideally here what we should have:

```bash
STORAGE_PROVIDER=dropbox DROPBOX_TOKEN=abc backup.sh januarybackup.tar.gz
```

`backup.sh` would:

* Create the `januarybackup.tar.gz` archive
* Mount `januarybackup.tar.gz` to `btcpay-dropbox` and run (note you do not need to explicitely pass the environment variables if your btcpay-dropbox image declare the variable with `ENV DROPBOX_TOKEN`)
* Delete `januarybackup.tar.gz` if succeed.

JeffVandrewJr:
@NicolasDorier I have implemented all of your suggestions and updated `backup.sh` accordingly to allow for custom filenames.

The only exception is that I did continue to pass the `DROPBOX_TOKEN` from the `docker run` command rather than through the Dockerfile. If I set `ENV DROPBOX_TOKEN=$DROPBOX_TOKEN` in the Dockerfile, it sets it from the CircleCI build environment.

astupidmoose:
Thinking of having this as an automated cron, Would it make more sense to change the filename="backup.tar.gz" to perhaps filename="backup_$(date "+%b-%d-%Y-%H-%M-%S").tar.gz" or something similar? This would ensure duplicate files are not overwritten, unless thats the goal?

JeffVandrewJr:
@astupidmoose 

The filename is only `backup.tar.gz` if you don't pass the script an argument. You could pass what you've got there as an argument when you write your crontab if you want something like that. 

At least with the Dropbox client, you'd probably want the files to be constantly overwritten because Dropbox automatically stores prior versions and allows you to restore them. The prior versions don't count against your storage alottment if you do it that way.

astupidmoose:
Yea, your right that probably makes more sense :)

JeffVandrewJr:
Changed the backup to the `/var/lib/docker/volumes` level rather than `/var/lib/docker`. Thanks to @astupidmoose for suggesting.

Kukks:
This is cool, once this is merged,  I will add a new provider to the bash script "BTCPAY", which will copy the backup file to a drop folder in the btcpay docker volume which in turn will trigger a file upload to any of the configured cloud providers. :D

NicolasDorier:
Will review and test it later today. It is super cool.

@JeffVandrewJr I think the Restore part should be handled by the same image though. (we can do later though, it should not be that complicated)

NicolasDorier:
Merging this, I need to properly document it though. @pavlenex might interest you.

JeffVandrewJr:
@NicolasDorier The backup tarballs this produces are uploaded unencrypted right now, as I wasn't sure how we wanted to handle the encryption step.

NicolasDorier:
unsure if it should be encrypted. This would be another piece of data (key, pass) that need to be saved and can be leaked and lost by the user.

PR #121, opened 2019-04-04
Add How BTCPay works video to docker readme

pavlenex:
Not sure if the position is appropriate, feel free to re-position if there's a better place.


NicolasDorier:
argh, it does not look like a clickable video

pavlenex:
@NicolasDorier good point, changed the thumbnail of the video to make the template more consistent with videos we have. Should be more obvious now.

PR #123, opened 2019-04-08
Introduce Exclude Fragments option

Kukks:
Allows you to exclude any fragments that get added. 
Good use case: do not deploy litecoin lightning node when using btc lightning

Should also help users with advanced customization scenarios


btcpay-setup.sh
```diff
@@ -66,6 +66,7 @@ Environment variables:
 Add-on specific variables:
     LIBREPATRON_HOST: If libre patron is activated with opt-add-librepatron, the hostname of your libre patron website (eg. librepatron.example.com)
     WOOCOMMERCE_HOST: If woocommerce is activated with opt-add-woocommerce, the hostname of your woocommerce website (eg. store.example.com)
+	BTCPAYGEN_EXCLUDE_FRAGMENTS:  Semicolon-separated list of fragments you want to forcefully exclude(eg. litecoin-clightning)
```
NicolasDorier: space not tab

PR #125, opened 2019-04-15
Add btc transmuter

Kukks:
Add Btc Transmuter to docker install


NicolasDorier:
Why `BTCTRANSMUTER_CRYPTOS` ? you can't use what the user already set up with BTCPAYGEN_CRYPTOX ?

Kukks:
Because the docker generator would need to become aware of the plugin to do
that. Did not want to pollute it with opt in services

NicolasDorier:
The container does not close nicely.

NicolasDorier:
What is this entrypoint ? 

https://github.com/btcpayserver/btcTransmuter/blob/master/BtcTransmuter/entrypoint.sh

Please either call dotnet process directly from dockerfile (like I did for NBXplorer) or use `exec dotnet ...` so that your process can see SIGKILL signals sent to it.

NicolasDorier:
`NBXplorer is not configured from the looks of it.`

NicolasDorier:
@Kukks need to limit the amount of stuff the user need to manually setup, else they will come in the chat and we lose time debugging them.

Maybe just build this parameter inside the `btcpay-setup.sh` from the BTCPAYGEN stuff.

Kukks:
> What is this entrypoint ?
> 
> https://github.com/btcpayserver/btcTransmuter/blob/master/BtcTransmuter/entrypoint.sh
> 
> Please either call dotnet process directly from dockerfile (like I did for NBXplorer) or use `exec dotnet ...` so that your process can see SIGKILL signals sent to it.

The entrypoint is so that I can restart the app from inside the container itself. This is so I can reload runtime plugins if I ever add support.
Will remove for now

Kukks:
> @Kukks need to limit the amount of stuff the user need to manually setup, else they will come in the chat and we lose time debugging them.
> 
> Maybe just build this parameter inside the `btcpay-setup.sh` from the BTCPAYGEN stuff.

I will create a generic env var that is built the generator that can be used

btcpay-setup.sh
```diff
@@ -141,6 +140,17 @@ if [[ "$BTCPAYGEN_REVERSEPROXY" == "nginx" ]] && [[ "$BTCPAY_HOST" ]]; then
     BTCPAY_HOST="$DOMAIN_NAME"
 fi
 
+BTCPAY_CRYPTOS=""
+for i in $BTCPAYGEN_CRYPTO1 $BTCPAYGEN_CRYPTO2 $BTCPAYGEN_CRYPTO3 $BTCPAYGEN_CRYPTO4 $BTCPAYGEN_CRYPTO5 $BTCPAYGEN_CRYPTO5 $BTCPAYGEN_CRYPTO6 $BTCPAYGEN_CRYPTO7 $BTCPAYGEN_CRYPTO8
+do  
+    echo "checkign loop $i";
```
NicolasDorier: typo

btcpay-setup.sh
```diff
@@ -141,6 +140,17 @@ if [[ "$BTCPAYGEN_REVERSEPROXY" == "nginx" ]] && [[ "$BTCPAY_HOST" ]]; then
     BTCPAY_HOST="$DOMAIN_NAME"
 fi
 
+BTCPAY_CRYPTOS=""
+for i in $BTCPAYGEN_CRYPTO1 $BTCPAYGEN_CRYPTO2 $BTCPAYGEN_CRYPTO3 $BTCPAYGEN_CRYPTO4 $BTCPAYGEN_CRYPTO5 $BTCPAYGEN_CRYPTO5 $BTCPAYGEN_CRYPTO6 $BTCPAYGEN_CRYPTO7 $BTCPAYGEN_CRYPTO8
```
NicolasDorier: should be in quote like `"$BTCPAYGEN_CRYPTO4"`

Kukks: done

btcpay-setup.sh
Kukks: this was outdated

NicolasDorier:
image

Kukks:
Merge this when you can @NicolasDorier  :)

NicolasDorier:
```
WARNING: The BTCPAY_CRYPTOS variable is not set. Defaulting to a blank string.
```

NicolasDorier:
Oh maybe it is because I run btcpay-update without setup.

PR #127, opened 2019-04-18
Add a simple template mechanism for fragments

Kukks:
This small change extends the generator to be able to replace placeholder test in fragments. Right now I'm only using it for an internal value (Cryptos, replaces #CRYPTOS# in the generated yml with the selected crypto codes e.g.: "BTC;LTC") but it could prove to be useful for other things later on.


NicolasDorier:
I still need to think about this.
I think for your specific problem on CRYPTOS you can just build the variable in the btcpay-setup.sh, this is the simplest.

Kukks:
Not needed anymore

PR #132, opened 2019-05-16
Add Eclair Support 

Kukks:


docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,55 @@
...
+      BITCOIN_EXTRA_ARGS: |
+        zmqpubrawblock=tcp://0.0.0.0:28332
+        zmqpubrawtx=tcp://0.0.0.0:28333
+        deprecatedrpc=signrawtransaction
+        addresstype=p2sh-segwit
+        rpcuser=ceiwHEbqWI83
+        rpcpassword=DwubwWsoo3
+  eclair_bitcoin:
+    restart: unless-stopped
+    stop_signal: SIGINT
```
NicolasDorier: is it really needed? Can you try if eclair close nicely without it?

Kukks: we dont need it, just copy pasta

NicolasDorier: tested. We actually do need it.

Kukks: Are you sure? It should exit gracefully?

Kukks: ah nvm, this was with my old custom image before

NicolasDorier: Actually you need SIGKILL.

Kukks: @araspitzu told me to use SIGINT instead of SIGKILL if anything

docker-compose-generator/docker-fragments/bitcoin.yml
```diff
@@ -11,7 +11,6 @@ services:
           rpcport=43782
           rpcbind=0.0.0.0:43782
           port=39388
-          disablewallet=1
```
NicolasDorier: Ok so this one is complicated... Old deployment of BTCPay had no `disablewallet`, so a wallet file got created.
Now if you remove this, Bitcoin core will want to rescan from the wallet that got created long ago to the tip. But if we run in pruned money then... the node won't start.

Instead, can you just put the bitcoin core wallet on a separate volume? Not only this prevent this problem but also it will make sure people don't delete their wallet by mistake when they only want to rescan their node.

Kukks: Is this even possible? The wallet is in a directory that is already used as a volume

NicolasDorier: you need to use `walletdir`. But actually the tricky part is that you need a different walletdir depending on the network...
So I think in the end you need to modify the bitcoin core image, so in the entrypoint, you create the folder for the user at the `WALLET_DIR/{network}` environment variable location. Then use walletdir= `WALLET_DIR/{network}`

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
araspitzu: SIGINT does terminate eclair and it's more graceful than SIGKILL, consider that it *should* also be safe to kill it as well but it's bad practice.

docker-compose-generator/docker-fragments/bitcoin.yml
Kukks: (ノ ゜Д゜)ノ ︵ ┻━┻

Ok will get on it

Kukks: https://github.com/NicolasDorier/docker-bitcoin/pull/21

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,56 @@
...
+    image: acinq/eclair:v0.3
+    environment:
+      PUBLIC_HOST: ${BTCPAY_HOST}
+      JAVA_OPTS: |
+        -Xmx256m
+        -Declair.printToConsole
+        -Declair.headless
```
araspitzu: Hey this is not necessary when running eclair-node, which is what acinq/eclair:v0.3 does.

Kukks: I think it's ok to be a little verbose here for a quick glance at settings :)

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,56 @@
...
+    links:
+      - bitcoind
+  btcpayserver:
+    environment:
+      BTCPAY_BTCLIGHTNING: "type=eclair;server=http://eclair_bitcoin:8080;password=DwubwWsoo3;bitcoin-host=bitcoind:43782;bitcoin-auth=ceiwHEbqWI83:DwubwWsoo3"
+    volumes:
+      - "eclair_bitcoin_datadir:/etc/eclair_bitcoin"
+    links:
+      - eclair_bitcoin
+      - bitcoind
```
Kukks: @NicolasDorier  Should I remove the direct dependency to bitcoind? We do not use the `GetDepositAddress` in btcpay right now and  would not need it if @araspitzu  implements https://github.com/ACINQ/eclair/issues/1006

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
araspitzu: Oh i referred to `-Declair.headless`, it has no effect when passed to eclair-node jar. That option is meant to run eclair-node-gui in headless mode but in the docker we only build eclair-node.

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
NicolasDorier: Yes, should remove this.

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,55 @@
...
+      - "9735" # server port
+      - "8080" # api port
+    volumes:
+      - "bitcoind_dir:/etc/bitcoin"
+      - "eclair_bitcoin_datadir:/data"
+    links:
+      - bitcoind
+  btcpayserver:
+    environment:
+      BTCPAY_BTCLIGHTNING: "type=eclair;server=http://eclair_bitcoin:8080;password=DwubwWsoo3;bitcoin-host=bitcoind:43782;bitcoin-auth=ceiwHEbqWI83:DwubwWsoo3"
```
NicolasDorier: remove bitcoind stuff.

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,55 @@
...
+services:
+  bitcoind:
+    environment:
+      BITCOIN_NETWORK: ${NBITCOIN_NETWORK:-regtest}
+      BITCOIN_EXTRA_ARGS: |
+        zmqpubrawblock=tcp://0.0.0.0:28332
+        zmqpubrawtx=tcp://0.0.0.0:28333
+        deprecatedrpc=signrawtransaction
+        addresstype=p2sh-segwit
+        rpcuser=ceiwHEbqWI83
```
NicolasDorier: remove this as well

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,55 @@
...
+
+services:
+  bitcoind:
+    environment:
+      BITCOIN_NETWORK: ${NBITCOIN_NETWORK:-regtest}
+      BITCOIN_EXTRA_ARGS: |
+        zmqpubrawblock=tcp://0.0.0.0:28332
+        zmqpubrawtx=tcp://0.0.0.0:28333
+        deprecatedrpc=signrawtransaction
+        addresstype=p2sh-segwit
```
NicolasDorier: do we really need this?

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
Kukks: Eclair needs it

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
Kukks: Eclair needs it

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
NicolasDorier: 1. eclair do not support cookie auth?
2. don't use rpcuser/password, instead use rpcauth. (like I did for lnd)
3. I don't know as I don't know if you can have more than 1 user per bitcoin instance, and one is created for lnd in the bitcoin image.

Kukks: @araspitzu can probably answer this better

NicolasDorier: If Eclair support cookiefile, it is really important that eclair actually read the file before making call, and do not cache the value, as bitcoind can restart and change the content of the cookiefile.

NicolasDorier: So I checked: Bitcoin core support multiple rpcauth.

So you just need to convert rpcuser/pass to rpcauth like I did in bitcoin-lnd (you need to pull, I just made new commit for this)

You can use NBitcoin for that RPCClient.GetRPCAuth or something like this.

araspitzu: Eclair only supports user/password authentication towards bitcoin, it also caches those values from the configuration so if you change them in bitcoin you need to update eclair.conf and restart eclair.

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
araspitzu: Line `deprecatedrpc=signrawtransaction` is not necessary anymore, since 0.3 eclair uses `signrawtransactionwithwallet`.

NicolasDorier: @araspitzu and about `addresstype`?

araspitzu: It can be removed as well, eclair supports both `p2sh-segwit` and `bech32` addresses and since bitcoin 0.16 the default is `p2sh-segwit` so this line can be safely removed.

NicolasDorier:
only missing, the RPCAuth

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,53 @@
+version: "3"
+
+services:
+  bitcoind:
+    environment:
+      BITCOIN_WALLETDIR: "/data/wallet"
```
NicolasDorier: This will be problematic if the user switch network, as the testnet wallet will be used for mainnet and vice versa, should depends on NBITCOIN_NETWORK.

Kukks: No, it creates a sub folder with the network type

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
Kukks: I switch to rpcauth

Kukks:
Will test it out again soon

araspitzu:
It's a bit sad this has stalled, is there anything i can do to help moving it forward?

Kukks:
I've been incredibly busy and used up all my test servers :( Will hopefully free one up again soon

Kukks:
If you can deploy this somewhere and give it a run through and see if there's anything missing would be best :)

NicolasDorier:
@araspitzu I heard that we can finally ask a deposit address from the eclair API? If so, we should bump version and use that instead. This will simplify things.

araspitzu:
Eclair doesn't offer an API for the deposit address, instead it's assumed that whoever controls eclair also controls the wallet provided by bitcoind. I'm not sure if I already mentioned it but if this is the only blocking issue than we're more likely to implement it to ease eclair-btcpayserver integration.

Kukks:
As far as I can remember, there is nothing blocking and everything is done
except for testing and iterative fixes. I think the biggest issue we will
see from users who install btcpay with eclair will be that their is no
remote UI available for eclair unlike the other implementations.

araspitzu:
Agreed, the lack of a web GUI will be a problem for the end users. Also eclair doesn't have the autopilot so if someone wants to use it with btcpayserver it must manually manage its channels via cmd-line (supported in eclair's docker) and that's not very user friendly. However this PR can be the first step toward full integration, if we make sure all the important bits work then when the web GUI arrives (yeah it's on my huge TODO list) it should be easy to add it.

NicolasDorier:
@araspitzu well then you should try to deploy a new btcpayserver based on kukks branch and see if it work

araspitzu:
I'm trying it right now but so far i couldn't get eclair up because of errors in the connection string, digging up the rabbit hole..

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,53 @@
...
+      BITCOIN_EXTRA_ARGS: |
+        zmqpubrawblock=tcp://0.0.0.0:28332
+        zmqpubrawtx=tcp://0.0.0.0:28333
+        rpcauth=eclair:d3a244200d5c946f6bef48e6583ec979$215049513aef33245dfec8ff3ab5e749458fde51ede931dec04c767c0732d468
```
araspitzu: Very subtle: there is a `$` sign causing docker-compose to fail parsing that, see [fix](https://github.com/araspitzu/btcpayserver-docker/commit/fb976a53dd4620d3593d99dce3a89cf280d9771d)

NicolasDorier: Indeed, needed that too for lnd.

Kukks: fixed

NicolasDorier:
@araspitzu please give another try and confirm if you can receive payment via eclair.

araspitzu:
Doesn't work for me: I see from the logs that eclair is started successfully but it's unreachable from btcpayserver. Most of the trouble is coming from the lightning connection string, if i select 'clicking here' from `You can use the internal lightning node by clicking here` in the `/lightning/BTC` view if fulfills the text area with `type=eclair;server=http://eclair_bitcoin:8080/;password=DwubwWsoo3;password=DwubwWsoo3` which has a duplicate password, if i remove one it will still consider it duplicate password (if i remove both it fails with Wrong Authentication). The `BTCPAY_BTCLIGHTNING` key in the docker files seem to be correct (also in the generated ones), i also tried removing the password from it but it gets stuck again in the `/lightning/BTC` view because either wrong auth or duplicate password. In fact i've not managed to get past that.

For completeness i'm starting `btcpay-setup.sh` with the following env variables:
```
NBITCOIN_NETWORK="regtest"
BTCPAYGEN_CRYPTO1="btc"
BTCPAYGEN_LIGHTNING="eclair"
BTCPAYGEN_ADDITIONAL_FRAGMENTS="bitcoin-eclair"
BTCPAY_HOST="localbtc.com"
```

NicolasDorier:
@araspitzu that's why I wanted a `getdepositaddress` from eclair's RPC. Because this is probably a problem of configuration of bitcoin RPC inside the connection string. Anyway, should be fixable.

araspitzu:
Oh alright, it's not clear to me the relationship between the eclair API password and bitcoind RPC authentication, from `bitcoin-eclair.yml` i see them having different [passwords](https://github.com/btcpayserver/btcpayserver-docker/pull/132/files#diff-9b985e5638df07d66bb4b77e504b55a9R34)

Kukks:
Actually this issue seems to be the same one I fixed in https://github.com/btcpayserver/BTCPayServer.Lightning/pull/16

I just tried locally with the same value for `BTCPAY_BTCLIGHTNING` and it doesnt cause this duplicate error

araspitzu:
It still puts two passwords in my instance, i'm working on [this branch](https://github.com/araspitzu/btcpayserver-docker/commits/eclair_integration) whose diff with this PR is the version of eclair docker image and the addition of `ports` section to expose LN protocol to the outside. I'm not sure it''s using master branch from https://github.com/btcpayserver/BTCPayServer.Lightning.

Kukks:
(ノ ゜Д゜)ノ ︵ ┻━┻
Let me try to rebase

Kukks:
Can you try again now?

araspitzu:
Okay the rebase did the trick and now i can connect btcpay to eclair! I have a few fixes that are necessary to run eclair, namely enabling `txindex` in bitcoind and exposing port `9735` to the outside, i will push them soon. However even with eclair set up and with channels the invoicing doesn't work, i paid an invoice via LN here on regtest but btcpay missed that and the invoice is still pending. I'm going to play around a bit more and will report back with feedback/logs.

Kukks:
Do you seriously need txindex though? 99% of deploys are pruned. While both LND and clightning state they require a non pruned node, we have managed to get it to work with pruning. We do wait for bitcoin to be fully synced before we run lnd/clightning. Would this work for eclair?

araspitzu:
Well, eclair relies on `getrawtransaction` and when bitcoin is not indexed that RPC only looks in the mempool which might not be enough. But it's a fair point, eclair _could_ get away without relying on `getrawtransaction` too much and shouldn't be a huge change, i'll investigate that but for the moment it's needed.

NicolasDorier:
@araspitzu try to make it works without `getrawtransaction` as @Kukks said, there is maybe less than 1% of our deployment being unpruned, and for the rest, they can't even be unpruned as it would increase their server cost by x2 or x3.

NicolasDorier:
And if possible, try to get a `getdepositaddress` at the same time :p

araspitzu:
The problem with pruning is that validating `channel_announcement`s becomes unreliable, the [`shortChannelId`](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#definition-of-short_channel_id) format requires eclair to get the block 'by height' calling `getblockhash <blockHeight>`  and depending on the pruning depth this might not work. A fair point is that a client should receive up-to-date `channel_announcements` of whom the blockHeight is not too far in the past but this is not guaranteed in any way. Getting rid of `getrawtransaction` is possible in eclair, the replacement would be `gettxout` with the addition that eclair needs to do some more bookkeeping to carry around the output index in addition to the transaction id, on the positive side this works without `txindex` and in pruned mode too. We're still evaluating if there are any undesired side effects and if it's worth doing this change, i will let you know.

araspitzu:
How does btcpay's LN node work with a pruned bitcoin instance? Every LN implementation needs to occasionally fetch arbitrary blocks. For example the first time you join the network there might be very old channels that are beyond your pruning point and you won't be able to validate them, as a result your local view of the network becomes defective and the ability to send payments is affected.

NicolasDorier:
@araspitzu because BTCPay is mainly for receiving money and not sending. Because the sender is the one deciding the route, most merchant does not care.

Kukks:
And even then, i have not had any issues with opening channels with anyone along with paying invoices through my 10 month old C-lightning on a pruned node

araspitzu:
Okay since receiving is more important for merchants the LN nodes on btcpayserver work fine even if the network view is incomplete. This best-effort approach makes sense and in the future with trampoline and rendez-vous payment the requirement of validating the entire graph will be even more optional so it makes sense to go in this direction. For the record i started changing eclair to not use `getrawtransaction` anymore and get rid of `txindex` i'm going push a PR soon-ish, this should work even in pruned mode - with the known quirks. One last remark: we updated the API recently and the work of @Kukks for the [eclair integration](https://github.com/btcpayserver/BTCPayServer.Lightning/pull/10)  is outdated now...so the latest docker image for `master` branch won't work with btcpay (the changes are mostly in the json model of the responses).

Kukks:
> 
> 
> Okay since receiving is more important for merchants the LN nodes on btcpayserver work fine even if the network view is incomplete. This best-effort approach makes sense and in the future with trampoline and rendez-vous payment the requirement of validating the entire graph will be even more optional so it makes sense to go in this direction. For the record i started changing eclair to not use `getrawtransaction` anymore and get rid of `txindex` i'm going push a PR soon-ish, this should work even in pruned mode - with the known quirks. One last remark: we updated the API recently and the work of @Kukks for the [eclair integration](https://github.com/btcpayserver/BTCPayServer.Lightning/pull/10) is outdated now...so the latest docker image for `master` branch won't work with btcpay (the changes are mostly in the json model of the responses).

That's a shitty last remark :D

araspitzu:
Indeed :D but it's only minor changes (they are necessary for the upcoming multi-part payments) and i can help you providing a digest of the diff, after i'm done with the `txindex` thing.

NicolasDorier:
while we are into breaking changes I wish for the `getdepositaddress` ! :D

araspitzu:
Update: we're working on [removing](https://github.com/ACINQ/eclair/pull/1178) the `txindex` requirement from eclair but there is no ETA at the moment and since it's a pretty big change it will take some serious testing before its released. Would it be okay to release an initial version with the `txindex` requirement and later on update?

Kukks:
Sounds good, can we move over to
https://github.com/btcpayserver/BTCPayServer.Lightning/issues so we can
adjust to your new API?

Kukks:
I mean https://github.com/btcpayserver/BTCPayServer.Lightning/issues/22

araspitzu:
I bumped eclair version to 0.3.2 and tried to pay a few invoices: btcpay did not notice the payment was successful.Do we have to rebase this on the latest from https://github.com/btcpayserver/BTCPayServer.Lightning ? I'm not sure if in my test i'm using the updated version with the fix for the new eclair API, @Kukks can you help me there?

Kukks:
Ah cool, you did the fix already.
We need to publish a new version of BTCPay.Ligthning, reference the new version in BTCPay, publish new version of BTCPay and then try here again

Kukks:
@NicolasDorier  can you bump btcpay.lightning? 

Once that's done, i'll publish a custom btcpay image so we can test it here :)

NicolasDorier:
@Kukks, no the tests with Eclair on BTCPayServer.Lightning still do not pass. I needed to disable eclair for it to pass. If the tests of the lib does not pass, this can't work on btcpayserver.

NicolasDorier:
I bumped btcpay.lightning with eclair tests passing and published a btcpayserver image with the new version. So if you rebase this PR, you can give it another try @araspitzu

araspitzu:
Tried it again and it works! The invoice was successfully paid on testnet lightning, it's good to note that i had to specify `BTCPAYGEN_ADDITIONAL_FRAGMENTS:bitcoin-eclair;opt-txindex` and i needed to open a channel via `bitcoin-eclair-cli.sh` then send out a few payments to have enough inbound liquidity to receive the payments, i'm not sure if [this](https://github.com/btcpayserver/btcpayserver-docker/pull/132/files#diff-9b985e5638df07d66bb4b77e504b55a9R38) is enough to be reachable for incoming channels, should we expose the port via `ports:`?

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
```diff
@@ -0,0 +1,53 @@
...
+      BITCOIN_WALLETDIR: "/data/wallet"
+      BITCOIN_EXTRA_ARGS: |
+        zmqpubrawblock=tcp://0.0.0.0:28332
+        zmqpubrawtx=tcp://0.0.0.0:28333
+        rpcauth=eclair:d3a244200d5c946f6bef48e6583ec979$$215049513aef33245dfec8ff3ab5e749458fde51ede931dec04c767c0732d468
+  eclair_bitcoin:
+    container_name: btcpayserver_eclair_bitcoin
+    restart: unless-stopped
+    stop_signal: SIGINT
+    image: acinq/eclair:eclair_cli_in_docker
```
araspitzu: I forgot to mention but this should be updated to `acinq/eclair:v0.3.2`, or if @NicolasDorier wants the `/getnewaddress` API i can tag a newer version and can use that one.

Kukks: updated. The `getnewaddress` part is not needed for initial release :)

NicolasDorier:
>  it's good to note that i had to specify BTCPAYGEN_ADDITIONAL_FRAGMENTS:bitcoin-eclair;opt-txindex 

Neither should be needed.

NicolasDorier:
I wish to get the getnewaddress for the release. I updated the lib, so I just need a properly tagged docker image, updating the library and release new btcpay version.

araspitzu:
Done! I pushed the tag `btcpay` and the image `acinq/eclair:btcpay` is now available on dockerhub.

NicolasDorier:
Cool. I think there will be a release in one or two days. Then can update this PR to remove the useless stuff in the connection string and we are good to go.

Next step will be multi arch image for docker! :p

NicolasDorier:
@araspitzu I noticed in the tests that it is possible to create several channel with the same node? Is it a bug or just something that only eclair support?

araspitzu:
The protocol doesn't forbid creating multiple channels with the same peer, IIRC `lnd` support it as well (only c-lightning doesn't).

Kukks:
C-lightning is the odd one here, with planned support down the line afaik

NicolasDorier:
@Kukks then there is a problem with the tests: this only happen with Eclair  https://circleci.com/gh/btcpayserver/BTCPayServer.Lightning/60

docker-compose-generator/docker-fragments/bitcoin-eclair.yml
NicolasDorier: Remove this, I will activate the wallet, but put it in a separate volumes so people who just need to resync won't dump their wallet by mistake....

NicolasDorier:
it is merged, closing.

NicolasDorier:
@araspitzu you should now try in on the master branch.
If it works, here are the work items to get more traction:

* Make multiarch image for docker so it works on small devices like raspberry
* Make it work with Tor

For both, you can look how we did in our own forks of lnd and clightning.

araspitzu:
Great! Tor is already supported in eclair, both v2 and v3 addresses, it should work out of the box but i need to give it a try with btcpay. The multiarch docker image is something i've been using for a while on the raspberry pi but i never proposed the change to be included in the official eclair, i will push for it because i know it's fairly important for btcpay users.

NicolasDorier:
@araspitzu also very important: make it works without txindex nobody has full blocks.

Issue #141, opened 2019-06-08
[question] Running with docker on machine where ports 80 and 443 aren’t free (without Tor)


woutersamaey:
I want to run BTCPay server on a machine that already runs a lot of docker containers and where ports 80 and 443 are already in use by an existing Nginx that I want to use to proxy/forward the traffic.

This Nginx forwards the traffic for all docker containers on the system.

How best to proceed?

NicolasDorier:
There is an option to run BTCPay without the NGINX server.

However, I advise you to not do it.

BTCpay need port 80 and 443 with NGINX because Let's encrypt (to get your certificate) require port 80 and 443.
Not only this, the NGINX in BTCPay is used heavily to redirect requests to the services inside the container. (The configuration is dynamic, depending of the services you use)

See https://vimeo.com/316630434 to understand better what it does.

woutersamaey:
I don’t mind using the built-in Nginx and having 2x Nginx. Only thing is mapping it on different ports. 

Let’s encrypt is already handled by my first Nginx and I’m terminating SSL there. 

I have done this in the past with BTC Pay Server and got it to work (the offloading).

So only 1 thing remains, how to change the ports on the built-in Nginx

NicolasDorier:
@woutersamaey the easier I see is that you fork btcpayserver-docker and you change the port mapping in `nginx.tmpl`.

The only downside is that you will need to rebase your changes on top of btcpayserver/btcpayserver-docker when you want to update. (though if you are lazy, you can just hack btcpay-update.sh to do it for your automatically)

woutersamaey:
Okay, I can see where you’re going with this. Any chance the ports can be configured using an environment variable? It would need to be passed from the bash script to the generation of the Nginx confit, but I’m not familiar with the template system to do it myself.

Willing to give it a shot though...

NicolasDorier:
@woutersamaey I think it can be good idea. Can you do a PR?

The template system should not scare you (see https://vimeo.com/316630434), it is actually very dumb.

NicolasDorier:
@woutersamaey note that you don't have to change the `.tmpl` of the nginx in your case. What you want to change is the port mapping in the `nginx.yml` (docker compose file), such that port 80 in container map to some other port on the host. 

This is super easy.

woutersamaey:
@NicolasDorier I've created the PR. Like you said. Super easy.

PR #142, opened 2019-06-08
New enviroment variables BTCPAY_HTTP_PORT and BTCPAY_HTTPS_PORT

woutersamaey:
@NicolasDorier as we discussed in #141 this is a PR for 2 new variables:
BTCPAY_HTTP_PORT
BTCPAY_HTTPS_PORT

I used these vars for Nginx (tested) and Traefic (not tested - I'm not familiar with it).

Ideas to make this better are:
- Option to enable SSL offloading & disable Let's Encrypt at the same time. I'm not sure if it is possible to disable whole sections in the nginx.yaml file. Maybe there should be 2 files (nginx.yaml and nginx-offloaded.yaml) and let a new env var BTCPAY_SSL_OFFLOADED choose between the 2?
- Improve documentation


NicolasDorier:
Cool! Will test. This is actually something that was asked to me quite a lot.

NicolasDorier:
So actually, this is not enough!
You need to add those settings in the `btcpay-setup.sh -i`, so it get saved in the profile.

woutersamaey:
Will do!

woutersamaey:
Any ideas on how to disable Let’s Encrypt? There’s a big fragment for it in Nginx.yaml. Is it possible to leave that entire thing out with just 1 var?

NicolasDorier:
@woutersamaey I will do later in another PR.

woutersamaey:
Another question?
Why does the setup script need “sudo” access (tested on Mac). Is it perhaps because it is needed to map port 80 and 443?

On Mac I know you need root access to bind ports < 1024, maybe other Linux systems as well. 

If this is the case, the need for sudo could be removed when using a port > 1024, which would also be very nice.

Will update my PR later today. It’s almost done, but have other Father’s Day things to do.

NicolasDorier:
@woutersamaey I am not against removing root requirement. However, don't do it as a default but as a parameter `--allow-no-root` or something like this.

By default, most user installing btcpay are not linux wizard and if the script run without root they would end up with a error message they can't understand.

woutersamaey:
Agreed, but what is the reason for needing root access? Is it this port mapping < 1024 or is there another reason? Or is it just for convenience and not sure?

Running docker as non-root is surely better (and for Mac users it is more practical as well since most Mac users have never done anything as root before).

NicolasDorier:
@woutersamaey because it need enough right to install docker, docker-compose, and run it.

NicolasDorier:
Maybe it is possible to run without root, but I don't have dedication to search about it.
Moreover, the container themselves run with root and I think we can't really rely on docker to be sure there is no way to escape from the chroot of the container.

woutersamaey:
Ok. I will put in the time to do some research and testing. Really prefer to run non-root as a security precaution.

woutersamaey:
@NicolasDorier can you please review this additional commit? I tried restarting the btcpayserver, rebooting the host, etc and can now see the new env vars being remembered, which was missing before.
Also added as much doc as I can and worded things as easily as possible.

NicolasDorier:
@woutersamaey I just split nginx.yml into an additional file: nginx-https which you can exclude via BTCPAYGEN_EXCLUDE_FRAGMENTS

btcpay-setup.sh
```diff
@@ -162,6 +165,8 @@ echo "
 Parameters passed:
 BTCPAY_PROTOCOL:$BTCPAY_PROTOCOL
 BTCPAY_HOST:$BTCPAY_HOST
+BTCPAY_HTTP_PORT:$BTCPAY_HTTP_PORT
+BTCPAY_HTTPS_PORT:$BTCPAY_HTTPS_PORT
```
NicolasDorier: Can you add
```bash
: "${BTCPAY_HTTP_PORT:=80}"
: "${BTCPAY_HTTPS_PORT:=443}"
```
Where I initialize other variable above?

btcpay-setup.sh
```diff
@@ -45,11 +45,14 @@ This script will:
...
 
-Make sure you own a domain with DNS record pointing to your website and that port 80 is accessible before running this script.
-This will be used to properly setup HTTPS via let's encrypt.
+Make sure you own a domain with DNS record pointing to your website.
+If you want HTTPS setup automatically with Let's Encrypt, leave BTCPAY_HTTP_PORT at it's default value of 80 and make sure this port is accessible from the internet.
+Or, if you want to offload SSL because you have an existing web proxy, change BTCPAY_HTTP_PORT to any port you want. You can then forward the traffic. Just don't forget to pass the X-Forwarded-Proto header.
 
 Environment variables:
     BTCPAY_HOST: The hostname of your website (eg. btcpay.example.com)
+    BTCPAY_HTTP_PORT: The port to bind to for public HTTP requests. Default: 80
```
NicolasDorier: Can you rename to
`REVERSEPROXY_HTTP_PORT`
`REVERSEPROXYS_HTTP_PORT`

woutersamaey:
Created a new commit with the 2 vars renamed and initialization. Hope this is okay @NicolasDorier ?

README.md
```diff
@@ -446,3 +449,17 @@ Then set it up:
...
+
+When you set `REVERSEPROXY_HTTP_PORT` to another value than 80, the built-in Let's Encrypt certificate will not work, as Let's Encrypt will try to validate your SSL certificate request by connecting from the internet to your domain on port 80. This validation request should be able to reach BTCPay Server in order to receive the certificate.
+
+If you need to run on a different port, it's best to terminate SSL using another web proxy and foreard your traffic. 
+
+## Can I offload HTTPS termination? 
+
+Yes. To offload SSL termination, just forward the requests to the port specified by `REVERSEPROXY_HTTP_PORT` and make sure you are setting the header `X-Forwarded-Proto: https` so BTC Pay Server can know the original request was HTTPS. If you forget this extra header, BTCPay Server will work, but it will believe the connection is insecure and display a warning message.
+
+Because you are offloading HTTPS, you won't need the built-in Let's Encrypt anymore and can exclude `nginx-https` by adding it to `BTCPAYGEN_EXCLUDE_FRAGMENTS`.
```
NicolasDorier: Add a command line example to do it.

README.md
```diff
@@ -446,3 +449,17 @@ Then set it up:
...
+
+You can change the ports for HTTP and HTTPS by setting the environment variables `REVERSEPROXY_HTTP_PORT` and `REVERSEPROXY_HTTPS_PORT`. This is handy when ports 80 and 443 are already in use on your host, or you want to offload SSL termination with an existing web proxy.
+
+When you set `REVERSEPROXY_HTTP_PORT` to another value than 80, the built-in Let's Encrypt certificate will not work, as Let's Encrypt will try to validate your SSL certificate request by connecting from the internet to your domain on port 80. This validation request should be able to reach BTCPay Server in order to receive the certificate.
+
+If you need to run on a different port, it's best to terminate SSL using another web proxy and foreard your traffic. 
+
+## Can I offload HTTPS termination? 
+
+Yes. To offload SSL termination, just forward the requests to the port specified by `REVERSEPROXY_HTTP_PORT` and make sure you are setting the header `X-Forwarded-Proto: https` so BTC Pay Server can know the original request was HTTPS. If you forget this extra header, BTCPay Server will work, but it will believe the connection is insecure and display a warning message.
```
NicolasDorier: Actually, this is not enough, see https://docs.btcpayserver.org/faq-and-common-issues/faq-deployment#btcpay-is-expecting-you-to-access-this-website-from

I advise you to link to the doc.

NicolasDorier:
I will test it, need change the doc (see my comments, you might actually have missed some stuff yourself)

README.md
woutersamaey: I read that doc, but IMHO there is no need to edit the original `/etc/nginx/conf.d/default.conf`.

The `map{}` directive isn't needed as you can use the built-in `$scheme` var instead.

This is the vhost config I currently use. Everything else is default Nginx on Debian Stretch.
```
cat /etc/nginx/sites-available/btcpayserver 

server {
	listen 80;

        root /var/www/html;
        index index.html index.htm index.nginx-debian.html;

        server_name yourdomain.com;

        location ~ /.well-known {
                allow all;
        }

        location / {
		return 301 https://yourdomain.com$request_uri;
        }
}

server {
        listen 443 ssl http2;

        ssl on;
        ssl_certificate      /etc/letsencrypt/live/yourdomain.com/fullchain.pem;
        ssl_certificate_key  /etc/letsencrypt/live/yourdomain.com/privkey.pem;

        root /var/www/html;
        index index.html index.htm index.nginx-debian.html;

        server_name yourdomain.com;

        location / {
                proxy_pass http://127.0.0.1:1080;

                proxy_set_header Host $http_host;
                proxy_set_header X-Forwarded-Proto $scheme;
                proxy_set_header X-Real-IP $remote_addr;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        }

        location ~ /.well-known {
                allow all;
        }
}
```

Agreed, more testing may be needed, but I'm able to access the site, backend, etc. Haven't done any heavy production use though.

woutersamaey: I'm not sure if these additional settings from the doc are actually needed.
Would be nice to see more comments on why they're there:
```
server_names_hash_bucket_size 128;
proxy_buffer_size          128k;
proxy_buffers              4 256k;
proxy_busy_buffers_size    256k;
http2_max_field_size       32k;
```

README.md
woutersamaey: Sorry, I don't follow. Using the env parameters is already explained, isn't it?

README.md
NicolasDorier: Sorry, yes.
So when you use the internal wallet, we pass PSBTs via request query `?psbt=blahblah`.
Problem is that if psbt is too long then you get a Connection Error.

I traced back this connection error on NGinx which serve HTTP2 content, even if the upstream is using HTTP. And I noticed those settings fixed the issue.

For `server_names_hash_bucket_size`, it is to allow long domain name without crash

README.md
NicolasDorier: Consider that people reading the doc, don't know how to edit an environment variable, can you provide them something to copy and paste?

README.md
woutersamaey: OK, this makes sense indeed. Can we update the page at https://docs.btcpayserver.org/faq-and-common-issues/faq-deployment#btcpay-is-expecting-you-to-access-this-website-from ?

The `map{}` part can be left out (it's a very confusing construction if you ask me, and not needed thanks to `$scheme`) and maybe include my vhost sample? It's cleaner.

So the user needs to do 2 things: copy-paste the proxy vars + create a vhost file in `sites-available` (always a good idea) + create a symlink to it in `sites-enabled`.

This doc section is for advanced users, so I don't think we should mind the extra complexity. Just follow best practises. The first time I read that doc, even I was confused by the single step to edit `/etc/nginx/conf.d/default.conf`. It just doesn't make sense if your Nginx is running more than 1 app.

NicolasDorier: Sure, please can you do a PR there you are more up to date to the best practices than I am.

NicolasDorier:
shit I merged my other request, can you fix the conflicts?

woutersamaey:
@NicolasDorier I rushed this a bit, so hope I didn't make matters worse...

NicolasDorier:
@woutersamaey yes, I moved some code in `helpers.sh` (the update of the docker env file), you forgot to save the ports there.

btcpay-setup.sh
```diff
@@ -45,11 +45,14 @@ This script will:
 You can run again this script if you desire to change your configuration.
 Except BTC and LTC, other crypto currencies are maintained by their own community. Run at your own risk.
 
-Make sure you own a domain with DNS record pointing to your website and that port 80 is accessible before running this script.
-This will be used to properly setup HTTPS via let's encrypt.
+Make sure you own a domain with DNS record pointing to your website.
+If you want HTTPS setup automatically with Let's Encrypt, leave REVERSEPROXY_HTTP_PORT at it's default value of 80 and make sure this port is accessible from the internet.
```
NicolasDorier: can you add back tilt around REVERSEPROXY_HTTP_PORT

btcpay-setup.sh
```diff
@@ -45,11 +45,14 @@ This script will:
...
+Or, if you want to offload SSL because you have an existing web proxy, change REVERSEPROXY_HTTP_PORT to any port you want. You can then forward the traffic. Just don't forget to pass the X-Forwarded-Proto header.
```
NicolasDorier: can you add back tilt around REVERSEPROXY_HTTP_PORT

woutersamaey: Is this okay?

NicolasDorier:
Still need to edit the `helpers.sh` to write in the docker env file.

NicolasDorier:
Can you fix that when you can? I will test again tomorrow.

woutersamaey:
will check when i get home

NicolasDorier:
I can test when you can.

woutersamaey:
Sorry for the delay (hectic times and big projects coming up).
I edited helpers.sh now.

NicolasDorier:
It works.

PR #143, opened 2019-06-08
bump transmuter to .24

Kukks:


NicolasDorier:
@Kukks when you make an update, instead you need to run the contrib/DockerBuilderblahblah/update-repo.ps1 script.

NicolasDorier:
This also automatically update the doc.

PR #164, opened 2019-07-05
 allow specification of additional domains with https support

Kukks:


btcpay-setup.sh
```diff
@@ -97,6 +97,7 @@ Environment variables:
     BTCPAYGEN_DOCKER_IMAGE: Allows you to specify a custom docker image for the generator (Default: btcpayserver/docker-compose-generator)
     BTCPAY_IMAGE: Allows you to specify the btcpayserver docker image to use over the default version. (Default: current stable version of btcpayserver)
     BTCPAY_PROTOCOL: Allows you to specify the external transport protocol of BTCPayServer. (Default: https)
+    BTCPAY_ADDITIONAL_HOSTS: Allows you to specify additional domains to your BTCPayServer with https support if enabled. (eg. example2.com,example3.com)
```
NicolasDorier: separated by `,` or `;`?

helpers.sh
```diff
@@ -61,6 +61,10 @@ btcpay_expand_variables() {
     if [[ "$BTCPAY_HOST" != *.local ]] && [[ "$BTCPAY_HOST" != *.lan ]]; then
         BTCPAY_ANNOUNCEABLE_HOST="$BTCPAY_HOST"
     fi
+    
+    if [ ! -z "$BTCPAY_ADDITIONAL_HOSTS" ]; then 
+        BTCPAY_ADDITIONAL_HOSTS=",${BTCPAY_ADDITIONAL_HOSTS// }"
```
NicolasDorier: what is `// }`

btcpay-setup.sh
Kukks: By `,` strictly. No spaces allowed or it malforms it so I sanitize any spaces in this further down

helpers.sh
Kukks: It removes any ` ` in the var
eg 
```
TEST="hello1world"
TEST="${TEST//1}"
echo "$TEST" //helloworld
```

NicolasDorier: Can you replace any `;` by `,` so that people don't complain if they messed up?

NicolasDorier: @Kukks you know what? I think I will just change the nginx template to separate on both `,` and `;` as well as ignoring empty entries so that you don't have to do this hack.

Kukks: Sure, should I remove from this pull request then?

NicolasDorier: no, just the hack part here.

NicolasDorier: @Kukks  can you remove this part and use 0.7.6 for docker-gen ?
Also change `,` to `;` in the nginx template.

docker-compose-generator/docker-fragments/btcpayserver-nginx.yml
```diff
@@ -7,10 +7,10 @@ services:
...
       VIRTUAL_NETWORK: nginx-proxy
       VIRTUAL_PORT: 49392
-      VIRTUAL_HOST: ${BTCPAY_HOST}
+      VIRTUAL_HOST: ${BTCPAY_HOST}${BTCPAY_ADDITIONAL_HOSTS}
       VIRTUAL_HOST_NAME: "btcpay"
       SSL_POLICY: Mozilla-Modern 
 
       # Let's encrypt settings
-      LETSENCRYPT_HOST: ${BTCPAY_HOST}
+      LETSENCRYPT_HOST: ${BTCPAY_HOST}${BTCPAY_ADDITIONAL_HOSTS}
```
NicolasDorier: add a `;` here

docker-compose-generator/docker-fragments/btcpayserver-nginx.yml
```diff
@@ -7,10 +7,10 @@ services:
       # NGINX settings
       VIRTUAL_NETWORK: nginx-proxy
       VIRTUAL_PORT: 49392
-      VIRTUAL_HOST: ${BTCPAY_HOST}
+      VIRTUAL_HOST: ${BTCPAY_HOST}${BTCPAY_ADDITIONAL_HOSTS}
```
NicolasDorier: add a `;` here

NicolasDorier:
Ok so: 
* Use docker-gen 0.7.6
* Keep using `,` and not `;`
* In the btcpay-setup.sh, throw an error if the user is using `;` because they will screw up.

NicolasDorier:
Then give a try on the host for hodlnaut and tor

Kukks:
@NicolasDorier  deployed with pr changes to https://fund.btcpayserver.org and https://weareallhodlonaut.com/

NicolasDorier:
coool!

NicolasDorier:
We need to document this in BTCPay server policies with link to doc.

pavlenex:
@NicolasDorier on it.

Kukks:
@pavlenex  You're up!
image

PR #168, opened 2019-07-14
Add opt-moar-memory to increase dbcache to 1GB in bitcoin.conf

jmaurice:
My new Raspberry Pi 4 with 4GB of ram laughs at your opt-save-memory setting


NicolasDorier:
I will rename it though, people will mess up the name for sure :p

PR #178, opened 2019-08-05
Add docker fragment for opt-txindex to enable txindex=1

wiz:


Kukks:
Also wondering if we should have a check to see if opt-save-storage is included and block installation if that's the case?

NicolasDorier:
@Kukks yes we should, can do it in the shell script.

wiz:
Okay I'll
1) Add documentation
2) Modify setup script to check for conflicting options

wiz:
k

PR #188, opened 2019-09-01
Add backdoor script to manually hack postgresql database

### LINKED ISSUE #187: Create an easy sql query bash script & simple utilities
We should add a bash script that can query the sql quicklyto handkle certasin scenarios from the docs easier

@wiz came up with this:
```
query='SELECT * FROM "U2FDevices"'
docker exec $(docker ps -a -q -f "name=postgres_1") su -c "psql -U postgres -d btcpayservermainnet -c '${query}'"
root@raspberrypi:/home/pi# ./foo.sh
```

we should create
`sql.sh <query>` - execute query
`reset-server-policy.sh` -> `DELETE FROM "Settings" WHERE "Id" = 'BTCPayServer.Services.PoliciesSettings';`
`set-user-admin.sh <email>`  -> `INSERT INTO "AspNetUserRoles" Values ( (SELECT "Id" FROM "AspNetUsers" WHERE "Email"='$1'), (SELECT "Id" FROM "AspNetRoles" WHERE "NormalizedName"='SERVERADMIN'));`
`disable-user-multifactory.sh <email>`  -> 
```
UPDATE public."AspNetUsers" SET "TwoFactorEnabled"=false WHERE upper('$1') = "NormalizedEmail";
DELETE FROM "U2FDevices" WHERE "ApplicationUserId" = (SELECT "Id" FROM "AspNetUsers" WHERE upper('$1') = "NormalizedEmail");
```

wiz:
Resolves #187

### DIFF
--- a/btcpay-admin.sh
+++ b/btcpay-admin.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+query()
+{
+    docker exec $(docker ps -a -q -f "name=postgres_1") psql -U postgres -d btcpayservermainnet -c "$*"
+}
+
+case "$1" in
+    reset-u2f)
+        query "SELECT * FROM \"U2FDevices\""
+        query "UPDATE public.\"AspNetUsers\" SET \"TwoFactorEnabled\"=false WHERE upper('\$1') = \"NormalizedEmail\""
+        query "SELECT * FROM \"U2FDevices\""
+        ;;
+    *)
+        echo "Usage: $0 [command]"
+        echo
+        echo "Commands:"
+        echo "         reset-u2f"
+        echo "         reset-u2f"
+esac
+
+exit 0


NicolasDorier:
coool, I think there should also have a way to set a user admin for when users lose their PWD

NicolasDorier:
added https://github.com/btcpayserver/btcpayserver-docker/commit/0dbcdd981ac53054d231f7f24ad203e5a8aec7f4

btcpay-admin.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+query()
+{
+    docker exec $(docker ps -a -q -f "name=postgres_1") psql -U postgres -d btcpayservermainnet -c "$*"
+}
+
+case "$1" in
+    reset-u2f)
+        query "SELECT * FROM \"U2FDevices\""
+        query "UPDATE public.\"AspNetUsers\" SET \"TwoFactorEnabled\"=false WHERE upper('\$1') = \"NormalizedEmail\""
```
NicolasDorier: I don't understand, whois `\$1` ?

Kukks: This is wrong, it should  be $2 so that user specifies which user should have its multifactor reset

NicolasDorier: I am unsure as the doc says there is no parameter.

wiz: I was using the SQL you gave me lol. Anyway yeah feel free to add or tweak whatever queries and other commands in the script, this was just to get you started 😅

PR #194, opened 2019-09-04
Add brittkelly utxo-set signature

britttttk:
Signing UTXO set


NicolasDorier:
You don't have to verify them all. You should only put the hash of the one that you actually verified.

britttttk:
Ok thanks please wait I will fix it.

PR #201, opened 2019-09-27
Extend backup.sh to allow backup using btcpay's storage provider

Kukks:
This uses the upcoming new feature in btcpay for an automatic storage drop folder https://github.com/btcpayserver/btcpayserver/pull/1054/files  to upload a backup of the btcpay data to any of the supported cloud storage providers.


NicolasDorier:
One moment, I wanted to cleanup the scripts to use `docker volume` commands instead of hard coding paths because extending this.

PR #203, opened 2019-09-30
xmr

Kukks:


docker-compose-generator/docker-fragments/monero.yml
```diff
@@ -0,0 +1,45 @@
+version: "3"
+
+services:
+  monerod:
+    restart: unless-stopped
+    container_name: btcpayserver_monerod
+    image: kukks/monero:v0.14.1.2
+    deploy:
```
NicolasDorier: remove this, depending on the docker-compose version of the host this will crash.

docker-compose-generator/docker-fragments/monero.yml
```diff
@@ -0,0 +1,45 @@
...
+  btcpayserver:
+    environment:
+      BTCPAY_CHAINS: "xmr"
+      BTCPAY_XMR_DAEMON_URI: http://monerod:18081
+      BTCPAY_XMR_WALLET_DAEMON_URI: http://monerod_wallet:18082
+      BTCPAY_XMR_WALLET_DAEMON_WALLETDIR: /root/xmr_wallet
+    volumes:
+      - "xmr_wallet:/root/xmr_wallet"
+
+
```
NicolasDorier: remove empty lines

docker-compose-generator/docker-fragments/monero.yml
Kukks: done

docker-compose-generator/docker-fragments/monero.yml
Kukks: done

NicolasDorier:
Cool. If the shitcoin does not work because it consumes too much resources, this is not the problem of btcpay.

Issue #216, opened 2019-11-03
Integrate ElectrumX Server - how to I make a pull request


maltokyo:
I made a custom fragment to integrate ElectrumX server to use my BTCPay full bitcoin node.  Thats cool, and works well.  Below is the content of my custom fragment.

I wanted to make a git pull request, but git confuses me, and despite my best attempts to do so watching tutorials etc, I could not work it out.  Anyway, sharing below, as this works well for me.

Content of custom fragment to enable ElectrumX Server to start at same time as BTCPay Server (and use its bitcoin full node):

```
version: "3"

services:
  electrumx:
    image: lukechilds/electrumx
    ports:
      - "50002:50002"
    volumes:
      - /path/to/somedrive/ElectrumX_data:/data
    environment:
      - DAEMON_URL=lnd:afixedpasswordbecauselndsuckswithcookiefile@bitcoind:43782
      - COIN=BitcoinSegwit
```

NicolasDorier:
Thanks for sharing, for someone who wants to make a formal PR:

You need to use a password specific to electrumx because the password used in this example is only available if you use lnd.

maltokyo:
Indeed. I didn’t workout how to make new PW so just stole the LND one.

What does “good first post” mean? ;)

NicolasDorier:
It means that if somebody feels like the motivation to make a Pull Request for this, it is a good issue not too complicated.

maltokyo:
OK, I think I worked out how to start that.  Cool, my first github pull request ;)
https://github.com/btcpayserver/btcpayserver-docker/pull/217

maltokyo:
Anyone willing to take a look at PR please do, from my side it's done and works well:
https://github.com/btcpayserver/btcpayserver-docker/pull/217

maltokyo:
Closing this, as PR is now in master and ready to use.

PR #217, opened 2019-11-03
ElectrumX Integration

maltokyo:
Use the full node of BTCPay Server with integrated ElectrumX from official source.  As per here:

https://github.com/btcpayserver/btcpayserver-docker/issues/216

This will work with LND active in BTCPay server, it needs modifying to work without LND as a standalone fragment (help would be appreciated).


NicolasDorier:
It is missing:

* Documentation in the README
* Using a different password

maltokyo:
Thank you @NicolasDorier  please see points below.

- Documentation in the README - I will update this
- Using a different password - tried myself to get this working, but could not, I need help from someone here
- Activate txindex in bitcoind - activating `opt-txindex.yml` in docker fragments needs to be pre-requisite.  I'll add in README.  I activated myself, but it only started indexing from block 500000 something.  Not sure why it didn't do the whole lot.  But, Electrumx Seems to be happy.

docker-compose-generator/docker-fragments/opt-add-electumx
```diff
@@ -0,0 +1,16 @@
+version: "3"
+
+services:
+  bitcoind:
+    environment:
+      BITCOIN_EXTRA_ARGS: |
+        txindex=1
```
maltokyo: txindex is mandatory for ElectrumX

docker-compose-generator/docker-fragments/opt-add-electumx
```diff
@@ -0,0 +1,16 @@
...
+  electrumx:
+    image: lukechilds/electrumx
+    ports:
+      - "50002:50002"
+    volumes:
+      - /path/to/somedrive/ElectrumX_data:/data
+    environment:
+      - DAEMON_URL=lnd:afixedpasswordbecauselndsuckswithcookiefile@bitcoind:43782
```
maltokyo: How to make a different working RPC password for another service?

NicolasDorier:
@maltokyo I am surprised that Bitcoind was happy, if it was off, it should have started reindexing from beginning. But the code looks fine.

Making a new password is a bit of a pain in the ass.

You need to use this utility: https://github.com/bitcoin/bitcoin/tree/master/share/rpcauth to know what to put in bitcoind params.

maltokyo:
@NicolasDorier thanks for the hint.  All done, PR should be good to go now.  Would you mind to review?

docker-compose-generator/docker-fragments/opt-add-electumx.yml
```diff
@@ -5,12 +5,13 @@ services:
     environment:
       BITCOIN_EXTRA_ARGS: |
         txindex=1
+        rpcauth=electrumx:7d8ee47c089e6072635f82b34796e878$13649d99453ccdf78e85007286422599c00e6953289f41bf8e92961076ba14db
```
NicolasDorier: you need to put an additional `$` (ie, `$$`) because the $ is interpreted by the yaml, you must escape it.

NicolasDorier:
Only the bug to fix, this is a good PR. 

However, after I merged this, I will remove the part where you expose electrumX. 

The traffic is not encrypted and authenticated, so this is dangerous, I will just expose via Tor instead.

maltokyo:
*@NicolasDorier 50002 uses SSL by default from ElectrumX it’s encrypted.
It’s absolutely needed to be exposed, otherwise Electrum wallets would not
be able to connect to it, rendering this whole thing completely useless.
Similar to how LND must expose 9735. *
*Whether or not people open their port 50002 to the Internet is up to them.
But it must be setup as I have it to be useful at least in local network.
Please reconsider this. *

NicolasDorier:
> *@NicolasDorier 50002 uses SSL by default from ElectrumX it’s encrypted.
It’s absolutely needed to be exposed

Cool then, do you have a test server I can try? 

>  otherwise Electrum wallets would not
be able to connect to it, rendering this whole thing completely useless.

They would, just over tor.

> Whether or not people open their port 50002 to the Internet is up to them.
But it must be setup as I have it to be useful at least in local network.

With Tor you don't have to open ports at all. But yeah maybe it is easier for local network.

maltokyo:
@NicolasDorier Thanks.  Updates as below:

1. "$" has been escaped, added another $ (thanks)
2. Would be good to open on TOR as well for sure, how can I do this?  50002 should be open also, and will not harm anything as it is designed to be open to the internet, and because most people will run this on their local network, and will not want to setup TOR just to connect their wallet to ElectrumX.
3.  Is naming conventions of this ok?:`container_name: btcpayserver_electrumx`
4. I added this, I assume ok: `restart: unless-stopped`
5. For the data directory, I changed to this, will it put into the same data dir as the rest of the BTCPay Server components?  Or is there something else needed?
```
    volumes:
      - "electrumx_datadir:/data"
```
6. You can test my setup (try from your side to connect from Electrum wallet) with my server: `94.16.175.196:50002` (50002 is open on my router and seems many peers connecting, so it should work, but I have never tried from outside my network!  So let me know).

NicolasDorier:
> "$" has been escaped, added another $ (thanks)

Don't worry for Tor, I can do it later that's quite simple.

> Would be good to open on TOR as well for sure, how can I do this?

Don't worry I can do it, that's simple.

> Is naming conventions of this ok?:container_name: btcpayserver_electrumx

You can remove this line actually.

> I added this, I assume ok: restart: unless-stopped

OK

> , will it put into the same data dir as the rest of the BTCPay 

To add it in the volumes section at the top level of the yaml file hierarchy. (check bitcoin.yml, at the bottom of the file)

> You can test my setup

I am trying but it does not work, this might come from me though... trying to find out.

maltokyo:
@NicolasDorier all done

1.
> You can remove this line `container_name: btcpayserver_electrumx` actually
 **removed**

2. TOR - thank you, would be helpful indeed to add

3.
> To add it in the volumes section at the top level of the yaml file hierarchy. (check bitcoin.yml, at the bottom of the file)
**Done** please check

4. Actually, from an external network, I try to test the connection to SSL port 50002 using this command `openssl s_client -connect 94.16.175.196:50002`, and it doesn't work.  Is there something else we need, to expose it in the same way as 9735 for LND and 443 for BTCPay?  All of 443, 9735 and 50002 are port forwarded on my router to the same host in the same way, but only 443 & 9735 seems to get through to BTCPay dockers.  What do I miss?  This is the error, so you definitely won't be able to connect until this connection is accepted:
```
# openssl s_client -connect 94.16.175.196:50002
139810203816384:error:0200206F:system library:connect:Connection refused:../crypto/bio/b_sock2.c:110:
139810203816384:error:2008A067:BIO routines:BIO_connect:connect error:../crypto/bio/b_sock2.c:111:
connect:errno=111
```

NicolasDorier:
btw, you can test your own pull request just by doing a `git checkout` to your own repo/branch and run `btcpay-update.sh`.

It would be good to know what is happening with this port. You configured your router NAT correctly?

maltokyo:
> You configured your router NAT correctly?
Yes, 50002 is set up for port forwarding exactly the same as 443, 9735 and those work from outside networks..  not sure what is happening.

> you can test your own pull request
Yes, I am running it now.  Everything works well, except being able to connect to 50002 from outside network as above.  Do I need to add anything into the network yml files, in order for IPTABLES to be set up correctly, etc?

maltokyo:
@NicolasDorier I spent really a long time, checking and triple checking ports on the router.
443, 9735 are available and working no problem from the outside world.  But 50002 is not.  Everything looks good in IPTABLES as below.  There must be some other docker networking setup config where I need to add port 50002 so that it is open to the outside world.. would you have some ideas?  I give up.  It works fine inside my network though, this is the mystery.

```
~# iptables -t nat -L -n
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination         
DOCKER     all  --  0.0.0.0/0            0.0.0.0/0            ADDRTYPE match dst-type LOCAL

Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         
DOCKER     all  --  0.0.0.0/0           !127.0.0.0/8          ADDRTYPE match dst-type LOCAL

Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination         
MASQUERADE  all  --  172.18.0.0/16        0.0.0.0/0           
MASQUERADE  all  --  172.17.0.0/16        0.0.0.0/0           
MASQUERADE  tcp  --  172.18.0.3           172.18.0.3           tcp dpt:50002
MASQUERADE  tcp  --  172.18.0.8           172.18.0.8           tcp dpt:39388
MASQUERADE  tcp  --  172.18.0.9           172.18.0.9           tcp dpt:9735
MASQUERADE  tcp  --  172.18.0.10          172.18.0.10          tcp dpt:443
MASQUERADE  tcp  --  172.18.0.10          172.18.0.10          tcp dpt:80

Chain DOCKER (2 references)
target     prot opt source               destination         
RETURN     all  --  0.0.0.0/0            0.0.0.0/0           
RETURN     all  --  0.0.0.0/0            0.0.0.0/0           
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:50002 to:172.18.0.3:50002
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:8333 to:172.18.0.8:39388
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:9735 to:172.18.0.9:9735
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:443 to:172.18.0.10:443
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:80 to:172.18.0.10:80
```

lukechilds:
@NicolasDorier 

>The traffic is not encrypted and authenticated, so this is dangerous, I will just expose via Tor instead.

A self signed SSL certificate is automatically generated on first run if one isn't provided: https://github.com/lukechilds/docker-electrumx/blob/43a81cf88fbd8997a159dad898e19c3d06e8c09b/bin/init#L3-L5

@maltokyo 

>When I started with txindex "on", it only started indexing from block 500000 something. Not sure why it didn't do the whole lot. But, Electrumx Seems to be happy.

After enabling `txindex` you'll need to restart with `bitcoind -reindex` to index the whole chain.

https://electrumx.readthedocs.io/en/latest/HOWTO.html#prerequisites

NicolasDorier:
@lukechilds @maltokyo well then this PR should be enough! I will test it tomorrow.

maltokyo:
Oh shit, 
I just restarted bitcoin node with `txindex=1` and afterwards added `reindex=1` too, to see what happens - it will take 48 days to index! so I removed the `reindex=1` and restarted again

now it starts to sync from 0%....  oops, that's a watch out..

NicolasDorier:
@maltokyo do you have a twitter?

maltokyo:
Yes ;) guess what it is ;)

maltokyo:
@maltokyo

NicolasDorier:
@maltokyo can you check if at least electrumx connect successfully to bitcoind by checking the logs of the container?

maltokyo:
Yes. It connected fine from the start. It even connected fine without full
tx index being available, as long as txindex = 1 is on on the node it
connected. I don’t see any outstanding issues.

NicolasDorier:
Awesome. We will make a blog post announcing all the cool stuff today.

NicolasDorier:
I added Tor support as well.

maltokyo:
thats fantastic!

PR #219, opened 2019-11-05
Expose Bitcoin RPC via Tor v3

NicolasDorier:


NicolasDorier:
Still need to create a fix login/pwd.

pavlenex:
@Fonta1n3 will be happy :-)

ChristopherA:
Some discussion about switching to use URI of by `standup:` instead. See: https://github.com/BlockchainCommons/Bitcoin-Standup#quick-connect-url-using-btcstandup

CC: @Fonta1n3

NicolasDorier:
I implemented using `btcrpc:` for now.

PR #258, opened 2020-01-20
Expose CLightning-Rest to BTCPay and external use

Kukks:
This should allow other lnd wallets to start supporting C-Lightning soon


Kukks:
tested over at ZeusLN by @kaloudis and confirmed to work. :)

PR #269, opened 2020-02-20
Add eclair for GRS and fix GRS LND

gruve-p:


NicolasDorier:
1. You need to `git update-index --chmod=+x` the two scripts
2. You need to modify `contrib/DockerFileBuildHelper/Program.cs`, then run the `update-repo.ps1` there, this will update the documentation with information on links of how people can build your eclair.

gruve-p:
@NicolasDorier ready for review

NicolasDorier:
You need to remove the chmod=+x on any other files than the two you added in this commit.

We noticed people having problem to update when permissions are changing like this. This is infortunate, but you can't fix the others now.

NicolasDorier:
Also you need to rebase on master, because the version of other docker are not the same than on master.

gruve-p:
@NicolasDorier Are all btcpayserver users having problems when permissions are changing or only the ones that are using grs scripts?

pavlenex:
A few users that reported that their updates were stuck due to GRS have never used GRS.

gruve-p:
What if we removed the faulty scripts now and added back the scripts at a later date like 3 months? Certainly there should be a solution to get the permissions right somehow?

gruve-p:
@NicolasDorier ready for review

NicolasDorier:
>What if we removed the faulty scripts now and added back the scripts at a later date like 3 months? Certainly there should be a solution to get the permissions right somehow?

Unsure, I don't even know how to replicate the issue.

PR #270, opened 2020-02-20
lnd-channels-remote-backup-on-change

bereska:
Currently BTCPay Server docker on every change in channels state saves static channels backup (SCB) locally on the same disk (on-site). This poses a threat to LN off-chain funds recovery in case of the server crash or fatal data loss.
Every time a change is made to channel.backup file in BTCPay Server docker it will instantly be copied to set up remotes.
This will allow you to keep the latest SCB offsite (on a remote server and/or popular cloud).
NOTE: this has been tested on Linux systems only.
TO BE CONTINUED ...)

### DIFF
--- a/lnd-channels-backup-dependencies.sh
+++ b/lnd-channels-backup-dependencies.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# run as root
+if [ "$(id -u)" != "0" ]; then
+    echo "This script must be run as root."
+    echo "Use the command 'sudo su -' and try again"
+    exit 1
+fi
+
+# generate LND backup ssh key
+if [ -f /root/.ssh/lnd_backup ]; then
+    echo "Key exists"
+else
+    echo "Generating SSH key"
+    ssh-keygen -o -a 100 -t ed25519 -f /root/.ssh/lnd_backup -N ''
+fi
+
+# check and install rsync and inotify
+echo "Checking rsync and inotify..."
+for pkgs in rsync inotify-tools; do
+        if [ $(dpkg -s $pkgs 2>/dev/null | grep -c "ok installed") -eq 1 ]; then
+                echo "$pkgs is already installed "
+        else
+                apt -yy install $pkgs
+                echo "Successfully installed $pkgs "
+        fi
+done
+
+# check and install rclone
+echo "Checking rclone..."
+if command -v rclone 2>/dev/null -eq 1; then
+        echo "rclone is already installed"
+else
+        curl https://rclone.org/install.sh | bash
+        echo "Successfully installed rclone"
+fi
--- a/lnd-channels-backup-systemd.sh
+++ b/lnd-channels-backup-systemd.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+sudo su -
+
+    echo "Adding backup-channels.service to systemd"
+    echo "
+[Service]
+ExecStart=/root/btcpayserver/lnd-channels-remote-backup-on-change.sh
+Restart=always
+RestartSec=1
+StandardOutput=syslog
+StandardError=syslog
+SyslogIdentifier=backup-channels
+User=root
+Group=root
+
+[Install]
+WantedBy=multi-user.target" > /etc/systemd/system/backup-channels.service
+
+systemctl daemon-reload
+systemctl start backup-channels
+systemctl enable backup-channels
+    echo "OK"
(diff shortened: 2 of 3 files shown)


Bereshka:
how cool ❤️

bereska:
@Kukks @NicolasDorier, the cloud backup of SCB from this PR can be integrated with #201 once the storage providers are set and are supported by rclone (currently more that 30). The rclone docker https://hub.docker.com/r/rclone/rclone would probably make more sense in this case. Let me know your thoughts

lnd-channels-backup-dependencies.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# run as root
+sudo su -
+
+# update system
+echo "Updating system"
+apt update && apt upgrade -y && apt autoremove
```
NicolasDorier: Remove this, thisis not the responsibility of this script.

lnd-channels-backup-dependencies.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# generate btcpayserver ssh key
+echo "Generating SSH key"
+ssh-keygen -o -a 100 -t ed25519 -f /root/.ssh/id_ed25519_btcpayserver -N ''
```
NicolasDorier: name the key better, like `lnd_backup`

lnd-channels-backup-dependencies.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# install inotify and rsync
+echo "Installing inotify and rsync"
+apt install -y inotify-tools rsync
```
NicolasDorier: don't install if already installed.

lnd-channels-backup-dependencies.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# install rclone
+echo "Installing rclone"
+curl https://rclone.org/install.sh | bash
```
NicolasDorier: don't install if already installed.

NicolasDorier: why not rsync?

bereska: rsync does the remote server part, rclone does the cloud 'cause it is suited for dealing with all these clouds filesystems, permissions, etc so much better

bereska: rclone has been well maintained, supports 30+ clouds and growing, is regular user-friendly with auto config, has built-in crypt option for advanced users

NicolasDorier:
This is a sweet idea, but I think it would be better if that was a docker image instead of separate script setup via systemctl.

The main reason is that we could then easily extend for monitoring other folder later, and it does not require modifications on the host.

bereska:
yes, I agree, already started working on it)
I take it to add a new container to this stack I need to create a new rclone.yml in docker-fragments and modify docker-compose.generated.yml accordingly. Is there a documentation I can look at to do it properly? Thank you for your guidance

NicolasDorier:
@bereska try to add it to the https://github.com/btcpayserver/dockerfile-deps repo.
You will need to make a docker file running your script above.
Take example on other dockerfiles, this one should not be that complex.

bereska:
@NicolasDorier , just fixed the scripts, working on dockerfile at https://github.com/btcpayserver/dockerfile-deps

bereska:
@NicolasDorier , just edited "Do not recreate if the key exists."

bereska:
@NicolasDorier , hope I didn't miss the point
https://github.com/btcpayserver/dockerfile-deps/pull/11

Kukks (changes requested):
Hi @Bereska, sorry it has taken so long with this PR. While I don't fully understand what this does, it would be better to group these scripts under  a directory like `Tools/lnd/backup`

@Bereska, do you think this script still works as-is?
@NicolasDorier  Did you have any other objections to this?
@rockstardev Maybe you know better regarding LND functionality?

bereska:
@Kukks hi, my script backs up to remotes lnd channels state on change. It's been successfully running on my rpi 4 with btcpay docker since 2019. But I failed to fully integrate it into docker file correctly as directed by @NicolasDorier

Goro2030:
This is a SO required feature. Why is it not being merged since December? Pleaaassseeee :)

NicolasDorier:
Closing this, old, pre AI, and undocumented. (where is it doing any backup)

PR #278, opened 2020-03-01
Fix PIHOLE_SERVERIP Variable and Add pi-hole as default Reverse Proxy Host

maltokyo:
Should be PIHOLE_SERVERIP="192.168.1.2" not PIHOLE_IPSERVER="192.168.1.2"

### DIFF
--- a/docs/pihole.md
+++ b/docs/pihole.md
@@ -25,7 +25,7 @@ From now everytime a device will connect to your local network, they will automa
 ## Using the dashboard
 
 Pi-Hole comes with a very nice admin dashboard to monitor its activity.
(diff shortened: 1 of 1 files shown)


peleion:
See my comments [here](https://github.com/btcpayserver/btcpayserver-docker/issues/279)

docs/pihole.md
```diff
@@ -51,6 +52,15 @@ Then running again
 docker logs pihole | grep random
 ```
 
+## Make pi-hole the eafult host for the Reverse Proxy Server
```
NicolasDorier: No, this is not recommended.

NicolasDorier:
Merging this, will remove the part about default host.

PR #282, opened 2020-03-10
update RTL

alevchuk:
closes btcpayserver/btcpayserver#1380


pavlenex:
Thanks for PR @alevchuk but we were a bit ahead of you and pumped to v.0.6.8, so I'll close this one. Welcome to make a new PR anytime when new versions are up.

NicolasDorier:
0.6.7 was not working.
Also it was missing some config changes that @Kukks already did.

Updated to 0.6.8 an hour ago with proper config.

PR #284, opened 2020-03-20
Add Electrumx low memory config

lukechilds:
This is untested but just saw the `opt-save-memory.yml` and thought I'd add some settings for Electrumx. It can be quite resource intensive, especially when it has a lot of connections.

Will these get correctly merged in to the `electrumx` block in the resulting `docker-compose.yml`?


maltokyo:
Hi @lukechilds yes this will work, when it's merged.  Thank you for adding!

NicolasDorier:
sorry for taking time, just saw this now.

PR #291, opened 2020-04-16
feat: integrate thunderhub

apotdevin:
# Feat: Integrate ThunderHub

-   Add nginx config to route `/thub` to thunderhub docker container
-   Add thunderhub docker build into docker compose
-   Update docs


README.md
```diff
@@ -331,6 +331,7 @@ We are trying to update our dependencies to run on `arm32v7` and `x64` boards. H
...
 | postgres | 9.6.5 | [✔️](https://raw.githubusercontent.com/docker-library/postgres/b7cb3c6eacea93be2259381033be3cc435649369/9.6/Dockerfile) | [✔️](https://raw.githubusercontent.com/docker-library/postgres/b7cb3c6eacea93be2259381033be3cc435649369/9.6/Dockerfile) | [✔️](https://raw.githubusercontent.com/docker-library/postgres/b7cb3c6eacea93be2259381033be3cc435649369/9.6/Dockerfile) | [Github](https://github.com/docker-library/postgres) - [DockerHub](https://hub.docker.com/_/postgres) |
+| apotdevin/thunderhub | 0.3.5 | [✔️](https://raw.githubusercontent.com/apotdevin/thunderhub/master/Dockerfile) | ️❌ | ️❌ | [Github](https://github.com/apotdevin/thunderhub) - [DockerHub](https://hub.docker.com/r/apotdevin/thunderhub) |
```
Kukks, line 334: Instead of adding this manually, can you also add the relevant build details to `contrib\DockerFileBuildHelper\Program.cs`
This file basically tells btcpayserver-docker how to automatically self build your docker image in case the user does not want to pull from DockerHub.

I think this would do
```cs
                case "bitcoin_thub":
                    dockerInfo.DockerFilePath = $"Dockerfile";
                    dockerInfo.GitLink = "https://github.com/apotdevin/thunderhub";
                    dockerInfo.GitRef = $"v{image.Tag.Split('-')[0]}";
                    dockerInfo.SupportedByUs = false;
                    break;
```

after adding that, run `update-repo.sh`

docker-compose-generator/docker-fragments/bitcoin-lnd.yml
```diff
@@ -57,6 +57,15 @@ services:
     expose:
       - "3000"
 
+  bitcoin_thub:
```
Kukks, line 60: Can you do a separate file for this - `opt-add-thunderhub.yml` and then include it in readme.md under `Available `BTCPAYGEN_ADDITIONAL_FRAGMENTS` currently are:`



At the bottom of the yml file, add 
```yml
required:
  - "bitcoin-lnd"
```

README.md
apotdevin: Will do, didn't know it was generated :D

apotdevin:
> Also, how you handle authentication? I'm guessing you will be exposing control of the lightning node.

To authenticate, users have to introduce their view only and admin macaroons which are stored in the browser. Admin macaroon is encrypted and needs the password for any admin required action.

README.md
apotdevin: Do I also need to add into `build-all-images.sh`? `update-repo-sh` gives me `Missing image info`

docker-compose-generator/docker-fragments/bitcoin-lnd.yml
apotdevin: Done! 👍

docker-compose-generator/docker-fragments/opt-add-thunderhub.yml
```diff
@@ -0,0 +1,12 @@
...
+services:
+    bitcoin_thub:
+        image: apotdevin/thunderhub:v0.3.5
+        restart: unless-stopped
+        environment:
+            BASE_PATH: "/thub"
+            HODL_KEY: "XQ86AOqw7Of8qMDql901mcZkuHL2nRel5FA9H2Ie4rTLeBKn0VlO5GuXvdVRogWrDOuH5V0E51Qxy9Bo"
+        expose:
+            - "3000"
+        required:
```
Kukks, line 11: this needs to be at the yml root level not on service level

docker-compose-generator/docker-fragments/bitcoin-lnd.yml
```diff
@@ -64,6 +64,7 @@ services:
       BTCPAY_BTCEXTERNALLNDGRPC: "server=/;macaroonfilepath=/etc/lnd_bitcoin/admin.macaroon;macaroondirectorypath=/etc/lnd_bitcoin"
       BTCPAY_BTCEXTERNALLNDREST: "server=/lnd-rest/btc/;macaroonfilepath=/etc/lnd_bitcoin/admin.macaroon;macaroondirectorypath=/etc/lnd_bitcoin"
       BTCPAY_BTCEXTERNALLNDSEEDBACKUP: "/etc/lnd_bitcoin/data/chain/bitcoin/${NBITCOIN_NETWORK:-regtest}/walletunlock.json"
+      BTCPAY_EXTERNALSERVICES: "Thunderhub:thub;"
```
Kukks, line 67: need to be in `opt-add-thunderhub.yml` not here
`

NicolasDorier:
You need LND 0.9.0 ? Right now BTCPayServer is incompatible with 0,9.0. There has been some breaking change in their version and nobody got time to find the issues and fix those.

apotdevin:
> You need LND 0.9.0 ? Right now BTCPayServer is incompatible with 0,9.0. There has been some breaking change in their version and nobody got time to find the issues and fix those.

No, doesnt need LND 0.9.0. With 0.9.0 you can do circular rebalancing but it is not necessary :)

NicolasDorier:
Kukks, seems good for you?
@apotdevin it would be nice to have multi arch images later. You can check how RTL does, that's not too complicated (or any other image we have)
Lot's of raspberry/rock64 nodes

contrib/DockerFileBuildHelper/Program.cs
```diff
@@ -544,6 +544,12 @@ private DockerInfo GetDockerInfo(Image image)
                     dockerInfo.GitRef = $"v{image.Tag.Split('-')[0]}";
                     dockerInfo.SupportedByUs = true;
                     break;
+                case "thunderhub":
+                    dockerInfo.DockerFilePath = $"Dockerfile";
+                    dockerInfo.GitLink = "https://github.com/apotdevin/thunderhub";
+                    dockerInfo.GitRef = $"{image.Tag.Split('-')[0]}";
```
Kukks, line 550: Are you sure about this? The git tags you have start with a `v`

apotdevin: The docker hub image also has the `v` so actually leaving it in gives me a double v and the following warning appears: `Broken link detected for image apotdevin/thunderhub:v0.3.6 (https://raw.githubusercontent.com/apotdevin/thunderhub/vv0.3.6/Dockerfile)`

Should work like this. Also don't know if it's good practice to have the docker hub image tag without the `v` but I just left it the same as github repo 😅

apotdevin:
> Kukks, seems good for you?
> @apotdevin it would be nice to have multi arch images later. You can check how RTL does, that's not too complicated (or any other image we have)
> Lot's of raspberry/rock64 nodes

Will have a look into it. Haven't done this before but will look into the other dockerfiles.

apotdevin:
Thanks! 🔥

apotdevin:
Hi! Just bumped the version, anything missing to get this merged? :)

Kukks:
Will merge soon, do you have an environment to test it afterwards?

apotdevin:
Hey, awesome. Yeah I have deployed instance to test this

Kukks:
Merged @apotdevin , let me know if it works ok

Kukks:
@apotdevin for next version, I recommend you 

* add a way to load macaroon from a provided directory ( see how rtl is configured here).
* add some way to auth: The preferred way is to generate a cookie file which a secret(password) and tell btcpay the location. BTCPay then loads it and generates a one-time url to access your app. Your app validates the secret against the file, and then generates a new cookie file(replaces old one) for the next auth attempt.
* add multi-arch images: to support raspberry pi, rock pi, etc

apotdevin:
Great, will test it out. Have added your suggestions to a feature request in ThunderHub repo. Regarding the second point wouldn't this mean that to access ThunderHub a new auth url would have to be created every time from BTCPay to access it? Considering sensitive information is stored locally in the browser (not taking into account if they can be loaded from a file) and encrypted is it really needed? But honestly unsure about the security issues here and what would be best.

Kukks:
> 
> 
> Great, will test it out. Have added your suggestions to a feature request in ThunderHub repo. Regarding the second point wouldn't this mean that to access ThunderHub a new auth url would have to be created every time from BTCPay to access it? Considering sensitive information is stored locally in the browser (not taking into account if they can be loaded from a file) and encrypted is it really needed? But honestly unsure about the security issues here and what would be best.

It's your call, it would basically allow you to automatically load the macaroon without asking the user + also leveraging BTCPay's authentication as protection. 

You can cache the auth result in a cookie and achieve the same result though.

apotdevin:
Can maybe add an account that requires this authentication but leave it open for other accounts with manually added macaroons. Then the user has the option to choose whatever he likes best.

Like the suggestion 👍 

Also, tested on a fresh deployment and is working as expected. Thanks for the pr review!

NicolasDorier:
@apotdevin can you make a tweet about it? we can retweet from btcpayserver account.

Kukks:
@NicolasDorier  https://twitter.com/tonyioi/status/1253233260018176002

apotdevin:
Thanks!

PR #298, opened 2020-04-26
change line to . ./btcpay-setup.sh -I line  177

Eskyee:
both Method works
.
. .
but this is the way describe in btcpay tooling scripts .

extra info for newbies
On Unix-like operating systems every directory contains, as a minimum, an object represented by a single dot and another represented by two successive dots. The former refers to the directory itself and the latter refers to its parent directory (i.e., the directory that contains it).


NicolasDorier:
no, not both method works. The one that you ask to replace assume you are in the right folder, not the current one.

Eskyee:
yes the game of linux dots jaja.

. The former refers to the directory itself. yes your correct.
. . parent directory (i.e., the directory that contains it).
 
I shall close this then. and thanks for clarification, Nicolas

PR #307, opened 2020-05-14
Add tip on changedomain.sh

RiccardoMasutti:
If you change the domain while U2F is active, you can no longer connect to the account.
This PR adds a little tip in the README.md file but I think it should also be shown inside BTCPay.

Can we create something like
`if` U2F is enabled
`then` display alert message/confirmation for changedomain.sh
?


pavlenex:
We do have documented process on how to reset [here ](https://docs.btcpayserver.org/faq-and-common-issues/faq-serversettings#how-to-disable-u-2-f-and-2fa-for-a-user). I'm unsure if adding a warring is critical, since let's be honest people set up domain and very rarely change it. If you want you can add a warring and link to doc.

NicolasDorier:
does not hurt to have a small notice

PR #310, opened 2020-05-19
chore: update and arm64

apotdevin:
-   Update to latest ThunderHub version
-   Add support for arm64v8 docker files


apotdevin:
The automatic generation of the `build-all-images.sh` file came with multiple changes. Should I revert the non related ones?

PR #317, opened 2020-06-08
chore: update thunderhub

apotdevin:


NicolasDorier:
@apotdevin does the cookiefile generation has been fixed? I have 2 PR to hopefully merge for next release to add thunderhub and only that was missing to work properly.

PR #321, opened 2020-06-17
Fix links in README

dennisreimann:
Found these while working on a broken link check for the docs.


README.md
```diff
@@ -140,7 +140,7 @@ A wide variety of useful scripts are available once BTCPay is installed:
 
 # Under the hood
 
-## Generated docker-compose <a id="generated-docker-compose"></a>
```
Kukks, line 143: why remove the anchor?

dennisreimann: Becuase the autogenerated one already is `generated-docker-compose`. When we leave the anchor in there it becomes `generated-docker-compose-` (trailing `-`, because the HTML gets replaced with a `-`)

PR #330, opened 2020-07-05
chore: bump thunderhub

apotdevin:
Unrelated but while running the update-repo script there seems to be a broken link for the docker image of `bitcore`


NicolasDorier:
thanks, I may remove it then.

PR #338, opened 2020-07-19
WIP: Add btc-rpc-explorer option

Kukks:
cc @janoside would be good if you started publishing official images

### DIFF
--- a/Production/nginx.tmpl
+++ b/Production/nginx.tmpl
@@ -45,6 +45,12 @@
 		proxy_pass http://bitcoin_thub:3000/;
 	}
 		{{ end }}
+		{{ if (eq $serviceName "btc_rpc_explorer") }}
+	location /btc-rpc-explorer/ {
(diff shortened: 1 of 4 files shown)


BitcoinMitchell:
Any change you/we could finish this @Kukks? This would def. be nice to have as that means that I don't have to worry about leaking data when I want to lookup some things.

Kukks:
I forgot what the issues were. Please feel free to pick it up or maybe @maltokyo  would add another awesome fragment to his portfolio :D

maltokyo:
Actually, I was thinking to add mempool.space - its better ;)
Shall we close this, and I will add mempool?

Kukks:
> Actually, I was thinking to add mempool.space - its better ;)
> Shall we close this, and I will add mempool?

image

maltokyo:
Done 
https://github.com/btcpayserver/btcpayserver-docker/pull/472

Issue #272, opened 2020-02-27
Migration of LND to seed backup does not work


hubiktomas:
Hello, I have tried to migrate my LND to seed backup. I am running BTCPayServer 1.0.3.161 deployed to Azure using the wizard about 2 years ago and just updating regularly since then. I have successfully closed all channels, waited for settlement, then run the command from the migration guide. Everything went smoothly, just the seed backup is not there - the instance is still telling me that the LND does not support seed backup and that I have to migrate. I tried to run the command 3 times including some reboots in between. Here is the output of the command:
```
root@BTCPayServerVM:/var/lib/waagent/custom-script/download/0/btcpayserver-docke                                                                                                                                                                                               r/Tools# ./recreate_bitcoin_lnd.sh
This script will delete and recreate your LND Bitcoin container. YOU CAN'T UNDO                                                                                                                                                                                                THIS OPERATION, ALL FUNDS THAT YOU CURRENTLY HAVE ON THIS LND WILL BE LOST! Type                                                                                                                                                                                                'yes' to proceed only after you've transfered all your funds from this LND inst                                                                                                                                                                                               ance
> yes
Only proceed if you've removed all the funds from LND Bitcoin container! This LN                                                                                                                                                                                               D instance will be completely deleted and all data from it unrecoverable. Type '                                                                                                                                                                                               yes' to proceed only if you are 100% sure
> yes
OK, last chance to abort. Type 'yes' to continue!
> yes
Stops containers and removes containers, networks, volumes, and images
created by `up`.

By default, the only things removed are:

- Containers for services defined in the Compose file
- Networks defined in the `networks` section of the Compose file
- The default network, if one is used

Networks and volumes defined as `external` are never removed.

Usage: down [options]

Options:
    --rmi type          Remove images. Type must be one of:
                        'all': Remove all images used by any service.
                        'local': Remove only images that don't have a custom tag
                        set by the `image` field.
    -v, --volumes       Remove named volumes declared in the `volumes` section
                        of the Compose file and anonymous volumes
                        attached to containers.
    --remove-orphans    Remove containers for services not defined in the
                        Compose file
Stopping letsencrypt-nginx-proxy-companion ... done
Stopping nginx-gen                         ... done
Stopping production_btcpayserver_1         ... done
Stopping nginx                             ... done
Stopping btcpayserver_lnd_bitcoin          ... done
Stopping btcpayserver_lnd_litecoin         ... done
[...]

PR #340, opened 2020-07-28
Deleting LND volume on old installations

rockstardev:
If volume is not present it won't throw any kind of error, it'll just echo volume name

image

Resolves: #272


NicolasDorier:
this would delete the LND install of old users, so not good.

PR #342, opened 2020-07-29
Revert "Deleting LND volume on old installations"

NicolasDorier:
Reverts btcpayserver/btcpayserver-docker#340

### DIFF
--- a/Tools/recreate_bitcoin_lnd.sh
+++ b/Tools/recreate_bitcoin_lnd.sh
@@ -21,10 +21,6 @@ fi
 
 docker volume rm --force generated_lnd_bitcoin_datadir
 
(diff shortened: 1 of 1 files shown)


NicolasDorier:
reverting it, if the volume does not work the script will fail @rockstardev

rockstardev:
No it won't... see the log here: https://github.com/btcpayserver/btcpayserver-docker/issues/272

And test it locally.

rockstardev:
@NicolasDorier see it executed:

deletion successful

NicolasDorier:
ok let's revert the revert

NicolasDorier:
reverted the revert.

PR #353, opened 2020-08-14
Add chatwoot support system

xpayserver:
Hi I add support for chat support system that soon integrate into BTCPay.


NicolasDorier:
wow that's cool. Will test it.

NicolasDorier:
Why requiring an environment file? You can set your own parameters with a custom docker fragment instead.

docs/chatwoot.md
```diff
@@ -0,0 +1,32 @@
...
+```bash
+CHATWOOT_HOST="chatwoot.xpayserver.com"
+BTCPAYGEN_ADDITIONAL_FRAGMENTS="$BTCPAYGEN_ADDITIONAL_FRAGMENTS;opt-add-chatwoot"
+. btcpay-setup.sh -i
+
+```
+4. Wait for BTPay to be online and then create the database for chatwoot
+
+```bash
+docker exec -ti chatwoot sh -c "export DISABLE_DATABASE_ENVIRONMENT_CHECK=1 && bundle exec rails db:reset"
```
NicolasDorier, line 29: Can't this step be done automatically with `entrypoint`?

xpayserver: no this only for create DB. Chatwoot no create db automatically..

xpayserver:
> Why requiring an environment file? You can set your own parameters with a custom docker fragment instead. This is how we do everywhere else.

This require because there many many config item that user must define which otherwise need pass through all btcpayserver-docker system. User need set SMTP, inbound email, storage which easy 20+ config https://github.com/chatwoot/chatwoot/blob/develop/.env.example

This PR work now but I switch focus to alt service #355  as it seem better. I merge both as both useful for user choice

NicolasDorier:
@xpayserver why not removing the env file and just documenting how to add a custom docker fragment to set all of those config?

britttttk:
@xpayserver can you post the github link so we can see what chatwood is?

Kukks (approved):
LGTM

uhlhosting:
Everyone want to look into erxes.io . I would see a way more easier integration, from paid or self hosted version managed via yarn, fast responsive development team.

PR #371, opened 2020-09-26
fix wrong parameter opt-unsafe-expose to opt-expose-unsafe in README.md

mlkui:
The parameter opt-unsafe-expose in README.md is wrong. It should be opt-expose-unsafe, otherwise an error will be shown:

	opt-unsafe-expose not found in /app/docker-fragments/opt-unsafe-expose.yml, ignoring...


pavlenex (approved):

pavlenex:
Awesome! 😄  Thanks for catching this up and congrats on your first contribution 🚀

PR #377, opened 2020-10-04
Added opt-expose-rpc.yml and related documentation.

max302:
I've been tooling around with Specter that requires RPC access to Bitcoind, I figured that other people might want/need this port opened.

I know this is highly unsafe, and people who can't figure this one out probably shouldn't be exposing RPC in the first place, but it might save a few minutes for somebody is technical but otherwise not familiar with the "assemble of fragments" model of configuration that BTCPayserver has.


NicolasDorier:
@max302 it is even a worse idea on btcpayserver which used fixed and well known password.
The only way to do it safe, is to use it over Tor (we already expose it by default)

max302:
The idea is to eventually integrate Specter inside the docker networking stack as with other modules, which I'm working on; right now the exposing of the RPC port only serves to test the daemon, which runs on the same machine. 

I'll reopen a PR when I have something workable.

NicolasDorier:
oh that would be awesome.

PR #386, opened 2020-10-26
Save $LIGHTNING_ALIAS env var to btcpay-env.sh

wiz:
Not sure why this isn't saved, but it seems LIGHTNING_ALIAS gets lost because it's not being saved to the profile.d/btcpay-env.sh file


Eskyee:
Yes can confirm as I seen this happen ..

Kukks (approved):

NicolasDorier:
This is saved in `cat $BTCPAY_ENV_FILE` file which is loaded.

Kukks:
Can confirm that my lightning alias is persisted without this PR. Will close for now but feel free to reopen @wiz  if you think there's still an issue.

PR #390, opened 2020-10-29
Add keysend to lnd via docker-fragment

cryptoteun:
Added dcoker-fragment to enable keysend functionality in Lightning.


NicolasDorier:
ping @rockstardev is there any reason to not enable it by default?

rockstardev:
Only if you don't want people sending you free BTC @NicolasDorier. There is no known attack vector, it just allows spontaneous payments - people being able to send you BTC using your pubkey.

I'm not sure if merchants would like this enabled by default though... considering they need to track every payment... and this allows anyone to send you BTC as long as they have your pubkey. So I guess we can keep it as opt-in fragment unless you prefer it enabled by default.

NicolasDorier:
@cryptoteun can you document it in the README?

cryptoteun:
@NicolasDorier, readme updated.

rockstardev (approved):
<3 @cryptoteun

PR #396, opened 2020-11-10
Add option to expose monerod 18081 (RPC port) on 127.0.0.1 on docker host

jooray:


docker-compose-generator/docker-fragments/opt-monero-expose.yml
```diff
@@ -0,0 +1,6 @@
+version: "3"
+
+services:
+  monerod:
+    ports:
+    - "127.0.0.1:18081:18081"
```
NicolasDorier, line 6: can you add

```yaml
required:
  - "monero"
```

So people who use this fragment without using monero get a proper error message?

Kukks (approved):

PR #410, opened 2020-12-09
bump dogecoin

astupidmoose:
This.. In theory.. will move dogecoin to the docker-deps repo instead of @rockstardev's fork.


rockstardev:
😢

PR #406, opened 2020-12-07
Wait for sync for C-lightning REST to start

Kukks:
Also bumps RTL and c-lightning-rest
fixes #384

PR #411, opened 2020-12-14
Removed Duplicates in update build-all-images.sh  the docker-gen generator needs to be double checked for bugs!!

Eskyee:
After digging around the codebase I Removed what I think are duplicates and  a old RTL version 
Other checks needed here…  Please!!  @rockstardev ..

Also I think the docker-gen generator needs to be double checked for bugs!! as this shouldn't happen..
 
please check these!!
# Build tor
# Build monero
# Build lnd

### DIFF
--- a/contrib/build-all-images.sh
+++ b/contrib/build-all-images.sh
@@ -153,22 +153,6 @@ docker build -f "$DOCKERFILE" -t "acinq/eclair:release-0.4.1" .
 cd - && cd ..
 
 
-# Build rtl
-# https://raw.githubusercontent.com/ShahanaFarooqui/RTL/v0.8.2/Dockerfile
-DOCKERFILE="Dockerfile"
-# https://raw.githubusercontent.com/ShahanaFarooqui/RTL/v0.8.2/Dockerfile.arm32v7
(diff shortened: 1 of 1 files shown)


NicolasDorier:
we generator the build-all-images with the docker with https://github.com/btcpayserver/btcpayserver-docker/tree/master/contrib/DockerFileBuildHelper so you need to dig around why it is making double there.

Eskyee:
I have a search 🔦 cheers

rockstardev:
@esky33kubernetes you can load csproj in your IDE and then run the project in debug mode. If you don't manage to debug it on your end, drop me a message on Mattermost and we'll schedule a call.

NicolasDorier:
I don't think those are duplicates, just different versions.

rockstardev:
After reviewing this in detail - I think @Eskyee is right. lnd has duplicate versions - regular LDN and groestlcoin LND.

But Tor and monero have same multiple entries. RTL has one duplicate entry and then old version. Since #406 is merged and RTL was bumped to 0.9.3 in all fragments, I'll update this file as well.

rockstardev (approved):

Eskyee:
thanks @rockstardev happy holidays .. We still grinding for 2021 xx Positive NRG`s

PR #412, opened 2020-12-14
bump monero

astupidmoose:


astupidmoose:
FYI, I've built the image locally based on the bumped docker-deps and its running without issue.

jooray:
> FYI, I've built the image locally based on the bumped docker-deps and its running without issue.

Same here, works. I don't yet know if it mitigates the height+2 attack, it needs to learn and block the nodes.

astupidmoose:
Closing this, as https://github.com/btcpayserver/btcpayserver-docker/pull/413 will apply a ban-list patch along with the bump and be better overall.

PR #415, opened 2020-12-22
Bump GRS Core and c-lightning

gruve-p:


Kukks (changes requested):
You need to change https://github.com/btcpayserver/btcpayserver-docker/blob/master/contrib/DockerFileBuildHelper/Program.cs#L466

Kukks (approved):

PR #416, opened 2020-12-25
Remove build-all-images.sh duplicates

rockstardev:
Continuation of #411, removing duplicate entries.


NicolasDorier:
this is useless, build-all-images is automatically generated by the update-repo.ps1 script.

Kukks (changes requested):
like @NicolasDorier  changing these files is pointless. We automatically generate these. If there is any bug, it is in the C# code that generates them

PR #419, opened 2021-01-01
Bump Monero version and migrate to DNSBL

sethsimmons:


wallopthecat:
this tag does not exist `     image: btcpayserver/monero:0.17.1.8`

jooray:
This broke btcpay-update.sh

Pulling monerod (btcpayserver/monero:0.17.1.8)...
ERROR: manifest for btcpayserver/monero:0.17.1.8 not found: manifest unknown: manifest unknown

sethsimmons:
I’ve updated https://github.com/btcpayserver/dockerfile-deps/blob/master/Monero/0.17.1.8/linuxamd64.Dockerfile, so I’m not sure if there is anything else I can do to get that image built via CI.

@NicolasDorier is there some other place that needs to be updated so the new tag is available in Docker Hub?

sethsimmons:
Opened https://github.com/btcpayserver/btcpayserver-docker/issues/420

NicolasDorier:
indeed fixed

PR #426, opened 2021-01-13
Memory and addnodes!

88plug:
- Create opt-addnodes.yml using Prohashing for faster sync (especially DOGE)
- Add DOGE to save and more memory
- Spelling error


88plug:
Need to fix         # Reducing memory usage of dogecoind. Don't try running this container without at least 2 GB of memory
        # https://www.reddit.com/r/dogecoin/comments/5wynqe/reducing_memory_usage_of_dogecoind/
        dbcache=50
Before pull request which should not be static

PR #427, opened 2021-01-13
DOGE RPC Security Fix + addnode fragment + memory fixes for DOGE

88plug:
DOGE is the only coin in BPS that uses a static rpcuser= and rpcpassword= that shouldn't be there.  
        rpcuser=ceiwHEbqWI83
        rpcpassword=DwubwWsoo3
is fair game for anyone to easily hack a new users node.

-Added unique rpcuser and rpcpassword generation for DOGE
-Removed static dbcache=50 so that the memory fragments will work properly (more-memory and save-memory)
-Addnodes fragment added using Prohashing explorer for eligable coins
-More memory and save memory now DOGE compatible
-Fixed spelling error

### DIFF
--- a/btcpay-setup.sh
+++ b/btcpay-setup.sh
@@ -333,7 +333,14 @@ if [[ "$NBITCOIN_NETWORK" != "mainnet" ]] && [[ "$NBITCOIN_NETWORK" != "testnet"
     echo "NBITCOIN_NETWORK should be equal to mainnet, testnet or regtest"
 fi
 
-
+if [[ $BTCPAY_CRYPTOS == *"doge"* ]]; then
+        echo "Found DOGE, generating unique RPC username and password for security"
+        random_user=$(date +%s | sha256sum | base64 | head -c 32 ; echo) ; sleep 1 ; random_pass=$(date +%s | sha256sum | base64 | head -c 32 ; echo)
+        perl -pi -e 's/rpcuser=.*/'"rpcuser=$random_user"'/g' docker-compose-generator/docker-fragments/dogecoin.yml
+        perl -pi -e 's/rpcpassword=.*/'"rpcpassword=$random_pass"'/g' docker-compose-generator/docker-fragments/dogecoin.yml
+        perl -pi -e 's/NBXPLORER_DOGERPCUSER: .*/'"NBXPLORER_DOGERPCUSER: $random_user"'/g' docker-compose-generator/docker-fragments/dogecoin.yml
+        perl -pi -e 's/NBXPLORER_DOGERPCPASSWORD: .*/'"NBXPLORER_DOGERPCPASSWORD: $random_pass"'/g' docker-compose-generator/docker-fragments/dogecoin.yml
+fi
 
 # Init the variables when a user log interactively
 touch "$BASH_PROFILE_SCRIPT"
--- a/docker-compose-generator/docker-fragments/opt-addnodes.yml
+++ b/docker-compose-generator/docker-fragments/opt-addnodes.yml
@@ -4,7 +4,7 @@ version: "3"
 services:
   dogecoind:
     environment:
-      BITCOIN_EXTRA_ARGS: |
+      DOGECOIN_EXTRA_ARGS: |
         addnode=coins.prohashing.com:6003
   feathercoind:
     environment:
--- a/docker-compose-generator/docker-fragments/opt-addnodes.yml
+++ b/docker-compose-generator/docker-fragments/opt-addnodes.yml
@@ -1,0 +1,26 @@
+version: "3"
+# Uses addnodes from Prohashing explorers for faster syncing! 
+
+services:
+  dogecoind:
+    environment:
+      BITCOIN_EXTRA_ARGS: |
+        addnode=coins.prohashing.com:6003
+  feathercoind:
+    environment:
+      BITCOIN_EXTRA_ARGS: |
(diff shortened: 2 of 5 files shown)


btcpay-setup.sh
```diff
@@ -333,7 +333,14 @@ if [[ "$NBITCOIN_NETWORK" != "mainnet" ]] && [[ "$NBITCOIN_NETWORK" != "testnet"
     echo "NBITCOIN_NETWORK should be equal to mainnet, testnet or regtest"
 fi
 
-
+if [[ $BTCPAY_CRYPTOS == *"doge"* ]]; then
```
Kukks, line 336: Please move this to a separate PR. I am unsure if there will be consensus around adding altcoin specific code in the main setup.

docker-compose-generator/docker-fragments/opt-addnodes.yml
```diff
@@ -0,0 +1,26 @@
...
+        addnode=coins.prohashing.com:6001
+  viacoind:
+    environment:
+      BITCOIN_EXTRA_ARGS: |
+        addnode=coins.prohashing.com:6095
+  monacoind:
+    environment:
+      BITCOIN_EXTRA_ARGS: |
+        addnode=coins.prohashing.com:6138
+exclusive:
```
Kukks, line 25: Remove this as there is no reason it would be exclusive since a node can have multiple addnode config items

docker-compose-generator/docker-fragments/opt-addnodes.yml
```diff
@@ -0,0 +1,26 @@
+version: "3"
```
Kukks, line 1: You need to document this in the README.

docker-compose-generator/docker-fragments/opt-save-memory.yml
```diff
@@ -16,6 +16,11 @@ services:
       BITCOIN_EXTRA_ARGS: |
         dbcache=50
         maxmempool=50
+  dogecoind:
+    environment:
+      DOGECOIN_EXTRA_ARGS: |
```
Kukks, line 21: Why does the node support less memory all of a sudden?

Kukks (changes requested):

btcpay-setup.sh
88plug: OK - will do - we need to do something though because static RPC user and pass will get users hacked eventually.

docker-compose-generator/docker-fragments/opt-addnodes.yml
88plug: OK

docker-compose-generator/docker-fragments/opt-addnodes.yml
88plug: Users should be able to add "addnodes" - depending on the use case.  Prohashing is a reliable source that runs seeder nodes, and I will be happy to document.

docker-compose-generator/docker-fragments/opt-save-memory.yml
Kukks: AFAIK save-memory for dogecoin was intentionally not introduced due to dogecoind needing all the ram

btcpay-setup.sh
Kukks: Not really, the dogecoin node is not exposed externally, it is firewalled inside the docker network. Many of the other services (lightning for example) have a hardcoded rpcauth in their yml file

docker-compose-generator/docker-fragments/opt-addnodes.yml
Kukks: I was referring to the `exclusive` section :)

docker-compose-generator/docker-fragments/opt-save-memory.yml
88plug: DOGE definetely does not need "all the ram" it needs a proper config though to manage it :)

docker-compose-generator/docker-fragments/opt-addnodes.yml
88plug: I thought exclusive was used when ignoring fragments, like this ```BTCPAYGEN_EXCLUDE_FRAGMENTS```
What is that section used for then?

btcpay-setup.sh
88plug: When DOGE is launched, Docker shows it is NOT using the "docker network"
```
0.0.0.0:22555
```
is exposed on the host as the rpcport!

Kukks: That's only inside the docker container. It is not the actual host machine

docker-compose-generator/docker-fragments/opt-addnodes.yml
Kukks: It is used to avoid conflicting fragments, for example someone including a pruning fragment with the txindex fragment

NicolasDorier:
We don't add alt specific settings to the setup. Fixed password are a problem we should fix, but this is not easy to do in a way that does not break existing installs.

uhlhosting:
When will this be pushed to any release? I feel the high usage on Doge chain sync and how slow it runs alto we run on SSD.

NicolasDorier:
Took the change from opt-*-memory on 51b700209b20babecb1114216dfa1b9b65e0284d kept the rest out.

PR #428, opened 2021-01-14
Added Prestashop as an opt

BitcoinMitchell:
I've tested this at least a bit locally and PrestaShop will be installed with the `BTCPay` moduled enabled by default. 

image

image

---

~~Blocked by: https://github.com/btcpayserver/dockerfile-deps/pull/24~~


BitcoinMitchell:
I currently do not have anything available to me which will allow me to test a full BTCPay Server setup with SSL, so I'm just going to assume it will work, as TOR seemed to work as well.

BitcoinMitchell:
I just realized, Tor will never work as PrestaShop will just redirect you to the host it has been set with. I'll remove that part now.

docker-compose-generator/docker-fragments/opt-add-tor.yml
```diff
@@ -41,6 +41,13 @@ services:
     volumes:
       - "tor_servicesdir:/var/lib/tor/hidden_services"
 
+  prestashop:
```
Kukks, line 44: Remove this if you don't want to expose over tor

NicolasDorier:
Don't worry too much about this. If that's not possible that's not too bad either.

Kukks (changes requested):

NicolasDorier:
```
PRESTASHOP_ADMIN_FOLDER
PRESTASHOP_PASSWORD
PRESTASHOP_EMAIL
```
Isn't it possible instead to have the user set that up the first time he starts prestashop?

docker-compose-generator/docker-fragments/opt-add-tor.yml
ping-localhost: Oh yes of course 🤦

BitcoinMitchell: Oh yes, of course 🤦

BitcoinMitchell:
> Isn't it possible instead to have the user set that up the first time he starts prestashop?

@NicolasDorier Let me try starting PrestaShop with `PS_INSTALL_AUTO: 0` and see if it asks me to do so. If not, then there is no other way.

BitcoinMitchell:
It is possible @NicolasDorier, we just need to merge https://github.com/btcpayserver/dockerfile-deps/pull/24 to ensure proper configuration.

---

The downside with the installer is that someone still has to remove the `install`-folder after they finished the setup. Which means they need to `exec` into the docker container. Not sure if that's really user friendly.

Kukks:
> It is possible @NicolasDorier, we just need to merge [btcpayserver/dockerfile-deps#24](https://github.com/btcpayserver/dockerfile-deps/pull/24) to ensure proper configuration.
> 
> The downside with the installer is that someone still has to remove the `install`-folder after they finished the setup. Which means they need to `exec` into the docker container. Not sure if that's really user friendly.

Hm, is there really no other way to do this? Maybe we can add a php script that executes after intsall in presta that deletes it?

BitcoinMitchell:
@kukks There is a post-install section (I see it in the docker image), but I can't find anything about that. I'll do some more research later on, but right now, I don't think we can get around that (unless we change the actual installer script to just remove it for us).

BitcoinMitchell:
I've looked and there is no easy or clean way around this. I highly recommend having a comment in the `README` over messing around with the PrestaShop source code and just hope that people will actually read what it says. 

It's quite clear and the README is filled with commands, so I assume it should be fine for most.

NicolasDorier:
@BitcoinMitchell made new image, will give it a try if we can remove it.

NicolasDorier:
Add `PRESTASHOP_HOST` in the `helpers.sh`

NicolasDorier:
There is some conflict with this mariadb and woocommerce one, I'll check if we can't just use the same database

NicolasDorier:
Ok so I tried it, first I needed to setup the store, which I think is OK if we document it.
However, once setup, prestashop redirect https request to http, but nginx redirect https to http, resulting in an infinite redirect loop.

Why would prestashop force using using http?

NicolasDorier:
You need to add `PS_ENABLE_SSL: 1`

NicolasDorier:
tried again, and I don't know how to make it work... even with the enable SSL, prestashop try to redirect to HTTP... :(

BitcoinMitchell:
> tried again, and I don't know how to make it work... even with the enable SSL, prestashop try to redirect to HTTP... :(
> The thing is that the install though is using HTTPS. So something inside prestashop try to use HTTP, this make no sense.
> 
> imho this is a bug of prestashop

When I tried it locally it also redirected me back to HTTP after setup, but in the [admin panel you can then enable SSL](https://doc.prestashop.com/display/PS17/General+parameters). We can add a comment telling people how to do this/add a command they can run that does it for them, but at that point it might just be easier to use the auto installer instead (as that will allow me to add post-install scripts).

BitcoinMitchell:
I just checked the PrestaShop source code and `PS_ENABLE_SSL` should set that database variables `PS_SSL_ENABLED` and `PS_SSL_ENABLED_EVERYWHERE` to true (relevant MR's: https://github.com/PrestaShop/docker/pull/152, https://github.com/PrestaShop/PrestaShop/pull/8668), which will enforce SSL (which I can confirm works locally, as I did update those variables by hand for testing). 

Could you do me a favor and fully remove the image + volume for Prestashop and the database and try again @NicolasDorier?

NicolasDorier:
I checked the source code, and the `PS_ENABLE_SSL` should start prestashop with `--ssl 1` which should set both `PS_ENABLE_SSL` and `PS_SSL_ENABLED_EVERYWHERE` to true.

When I did my test, I deleted prestashop volume and mariadb volume but it did not worked.

BitcoinMitchell:
I did try it, but didn't pay close attention to that final part. I'm going to close this PR for now and re-open once I have something actually usable.

PR #429, opened 2021-01-15
Improve and document backup script, add scp option

dennisreimann:
This should be backwards compatible, but I could not test the Dropbox option. 
Would be good if someone with a Dropbox account could check this.

Local and SCP backup options are working fine for me.


dennisreimann:
Did anyone test this?

PR #432, opened 2021-01-20
Docker fragment for Bitcoin Wallet Tracker

shesek:
Integration for [Bitcoin Wallet Tracker](https://github.com/bwt-dev/bwt), a lightweight descriptor-based wallet indexer similar to EPS, available as an Electrum RPC server and an HTTP API.

At minimum, this requires configuring one of `BWT_XPUBS` or `BWT_DESCRIPTORS` with the wallets to track. It is also recommend to set `BWT_RESCAN_SINCE` to the wallet creation date (as `YYYY-MM-DD`) to reduce rescanning time. The other options are [documented here](https://github.com/bwt-dev/libbwt#config-options), but users typically wouldn't need to change them.

I used the electrum-only docker image variant by default, which is smaller and comes with less dependencies. Enabling the HTTP API  requires setting `BWT_VARIANT=full`.

The ports are exposed locally and over Tor only, since the Electrum protocol is unauthenticated and could leak private information if exposed to the internet. (though I am planning to [hack around this limitation](https://github.com/spesmilo/electrum/issues/6778#issuecomment-748882815) in a future release.)

TODO:
- [ ] ~~TLS support (terminated with nginx)~~
- [x] Add multi-arch support to bwt's docker images
- [x] Document
- [x] Add env vars to `helpers.sh`
- [x] Add to `DockerFileBuildHelper`


shesek:
Do you think this should use the default bitcoind wallet by default, or create a separate one designated for the addresses imported by bwt?

If the user previously had just one wallet, one annoying side-effect of creating another one is that it makes it necessary to start specifying the wallet name when calling wallet-related rpc methods.

This can already be done by setting `BWT_BITCOIND_WALLET=bwt BWT_CREATE_WALLET_IF_MISSING=1`, just wondering what the default behavior should be.

Kukks (changes requested):
* I don't see the xpub/descriptors vars in the fragment.
* If the user can/must  specify env vars, you'll need to add them to the btcpay shell/ps1 scripts so that they are correctly loaded by btcpay between restarts, ssh sessions etc
* Also need to add docs to the README, specifying the fragment name, description and what vars are possible for it
* We allow users to specify that they want to build ALL the docker images that are used locally. Requires a few trivial lines here https://github.com/btcpayserver/btcpayserver-docker/blob/master/contrib/DockerFileBuildHelper/Program.cs#L564   and then running https://github.com/btcpayserver/btcpayserver-docker/blob/master/contrib/DockerFileBuildHelper/update-repo.sh

shesek:
> I don't see the xpub/descriptors vars in the fragment.

They don't have a default value so they don't appear there, but they work if they're set externally.

> If the user can/must specify env vars, you'll need to 

Yes, configuring the xpubs/descriptors currently requires setting env vars, though I'm planning to write a web UI for this at a later stage. Will do what you suggested.

> Also need to add docs to the README ...

Will do. The full list of options is kinda long, I guess I should probably stick to just the common ones?

> We allow users to specify that they want to build ALL the docker images that are used locally. 

That's pretty cool, I didn't realize btcpay enabled doing this in a streamlined manner. Will look into doing the necessary changes.

shesek:
I added some documentation in https://github.com/btcpayserver/btcpayserver-docker/pull/432/commits/a3939b05658b69f6ba9d97e5ef6ad3463dfa844c and modified `helpers.sh` to persist bwt's env vars.

Because bwt supports wildcard `BWT_XPUB_*`/`BWT_DESCRIPTORS_*` env vars for configuring multiple xpubs/descriptors (recently added, still unreleased), I made it save all the options beginning with `BWT_`. Is that ok? Something similar could also be done for the `BTCPAY_*`, `TOR_*` and `REVERSEPROXY_*` options to keep things shorter.

I [wasn't sure](https://chat.btcpayserver.org/btcpayserver/pl/1nuex4fj9t845x96b6js515bka) if I should also modify the `build.sh`/`build.ps1` files or not. I didn't for now.

shesek:
I updated the DockerFileBuildHelper. It didn't play nicely with the dynamic image name I was using, so I removed it in favor of only supporting the Electrum variant, which is what users are likely to want anyway. I'll revisit enabling the HTTP API later.

I ran the `update-repo.sh` script, but it re-ordered the images and resulted in a huge, difficult-to-audit diff, so I wasn't sure if I should commit it. Should I or will you?

shesek:
Updated to bwt v0.2.2 with multi-arch docker images.

SSL support isn't strictly necessary since this is only exposed over Tor, so I'm gonna put this aside for now.

Is there anything else still missing?

NicolasDorier:
@shesek for the `update-repo.sh`, as long as it does not crash that's ok.

NicolasDorier:
Ping @shesek for some obscure reason your PR broke our build (see https://github.com/btcpayserver/btcpayserver-docker/commits/master)

https://github.com/btcpayserver/btcpayserver-docker/commit/b9ffc73015072bc21a060aae3a1c9c8e9ad33633 fixed the build.

This however probably break bwt.

PR #433, opened 2021-01-23
Update DockerFileBuildHelper instructions

shesek:
Added instructions for building with Docker and for updating the repository.

I also removed a sentence that was cut midway through, *"Note that the"*. Did you mean to say anything important there? :)


Kukks (approved):
LGTM

Issue #392, opened 2020-11-02
Add clboss as fragment


RiccardoMasutti:
CLBOSS is an automated manager for C-Lightning forwarding nodes. It helps node operators.
https://github.com/ZmnSCPxj/clboss

max302:
This is a duplicate of #393, maybe delete or merge somehow?

RiccardoMasutti:
> This is a duplicate of #393, maybe delete or merge somehow?

These are two different plugins

PR #434, opened 2021-01-23
Add plugins dir volume to btcpay

Kukks:


Kukks:
@ndeet  this should fix the weird issue of the installed plugin disappearing between restarts

RiccardoMasutti:
Is it possible to use it to enable c-lightning plugins like #392?

thiagolcks:
Is there any documentation explaining how to use it to install a c-lightning plugin?

ksdhans:
> Is it possible to use it to enable c-lightning plugins like #392?

I'd like to know this too. I've just had almost all my inbound liquidity wiped out. Some channels were closed, while others are currently disconnected. I'd very much like to use something like the clboss plugin to manage this automatically.

PR #439, opened 2021-02-07
Bumping LND to v0.12.0-beta

rockstardev:
We need to keep using our https://github.com/btcpayserver/lnd fork because of nginx and Tor we need to register listener on 0.0.0.0.

@NicolasDorier let me know if you want to jump on refactoring the way we hit LND so that all other services start reaching it on `lnd_bitcoin:8080`.

Note: putting it in draft so you don't merge it until we talk about details.


NicolasDorier:
All services reach LND already on `lnd_bitcoin:8080`.
The question is why LND does not seems to listen on `lnd_bitcoin:8080` despite binding to it. I would have expected it to work.

NicolasDorier:
you might try to use netstat on the lnd container when biding to lnd_bitcoin to see at the OS level if LND is binding correctly or doing funny magic.

uhlhosting:
/root/.lnd/lnd.conf:13: unknown option: notls
When upgrading to 1.12.0 beta.

uhlhosting:
Seems to break thunderhub integration.

rockstardev:
@uhlhosting thanks for test, I'll open new PR with 0.12.1-beta and we try with fix they published.

rockstardev:
Closed this in favor of #450

PR #457, opened 2021-04-02
Separate default liquid prune to own fragment

Kukks:
This allows users who dont want to prune their liquid node to exclude the pruning fragment. (allows rescanning for utxos)

### DIFF
--- a/docker-compose-generator/docker-fragments/liquid-default-prune.yml
+++ b/docker-compose-generator/docker-fragments/liquid-default-prune.yml
@@ -1,0 +1,7 @@
+version: "3"
+
+services:
+  elementsd_liquid:
+    environment:
+      ELEMENTS_EXTRA_ARGS: |
+        prune=5000
--- a/docker-compose-generator/docker-fragments/liquid.yml
+++ b/docker-compose-generator/docker-fragments/liquid.yml
@@ -16,7 +16,6 @@ services:
         port=39388
         whitelist=0.0.0.0/0
         validatepegin=0
-        prune=5000
         fallbackfee=0.000001
     expose:
       - "43782"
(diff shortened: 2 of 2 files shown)


maltokyo:
How big is the full liquid chain?
I am running it, but only now realise I run in pruned mode

Kukks:
> How big is the full liquid chain?
> I am running it, but only now realise I run in pruned mode

Not that big (10.8GB for the full data folder, fully synced with txindex=1), the main concern was that users would attempt to enable it on their existing Lunanode installs and have their servers full up on space and crash

maltokyo:
Ah, I have fully synched and pruned.  Should I just delete the volume and start again to get full tx index, (after your split in this PR goes in), or can I do that somehow retrospectively?

Kukks:
> Ah, I have fully synched and pruned. Should I just delete the volume and start again to get full tx index, (after your split in this PR goes in), or can I do that somehow retrospectively?

Be careful of the wallet dir. BTCPay uses the elementsd wallet since we do not have a functional wallet solution yet for liquid(and for unblinding confidential assets).

Kukks:
Yes, deleting the dir should be enough (though my or does not add in txindex, it would need to be a separate fragment)

maltokyo:
cool, thanks!  I am also trying to work out how to do a peg-in on my own server.  Unrelated, but cannot find any info on how to do that, other than an old issue where you documented a bunch of stuff.

Kukks:
> cool, thanks! I am also trying to work out how to do a peg-in on my own server. Unrelated, but cannot find any info on how to do that, other than an old issue where you documented a bunch of stuff.

need to look into that but it will require hooking liquid to the btc node and some additional configs (and probably not be able to run btc in pruned mode)

maltokyo:
Ah, ok.  That's why I enabled it, to do my on peg-ins (such a weird word, has strange connotations..), or at least try.
I do have a full tx index and non-pruned node, so I can test if you find out.

PR #465, opened 2021-04-27
Add Taproot Client Opt In Option

Kukks:


Tools/recreate_bitcoin_lnd.sh
```diff
@@ -2,7 +2,18 @@
...
+ read -p "The unofficial taproot node release is already active `echo $'\n> '`" yn
+ if [ $yn != "yes" ]; then
+ 	exit 0
+ fi
+
+fi
+
+
+
+read -p "This script will  THIS LND WILL BE LOST! Type 'yes' to proceed only after you've transfered all your funds from this LND instance `echo $'\n> '`" yn
```
Zaxounette, line 16: This warning message is incomplete compared to the original.

Kukks: Thanks, fixed!

Tools/toggle_bitcoin_taproot_bip8_lottrue..sh
```diff
@@ -0,0 +1,39 @@
...
+  exit 0  
+fi
+
+
+
+read -p "This script will swap the official Bitcoin release with an unofficial version provided from https://bitcointaproot.cc that includes taproot activation signalling via BIP8 wiht LOT=TRUE. Type 'unofficial' to switch to this UNOFFICIAL FORK OF BITCOIN CORE. `echo $'\n> '`" yn
+if [ $yn != "unofficial" ]; then
+	exit 0
+fi
+read -p "OK, Are you sure you know what you are doing? Please be aware that you running a Bitcoin node that is not from the official release group and cycle. There can be a potential chain split in special circumstances. ONLY RUN THIS IF YOU ARE KNOWLEDGEABLE. BTCPAYSERVER DOES NOT PROVIDE AN OPINION ON THIS DECISION AND THIS OPTION DOES NOT MEAN WE ARE PROMOTING THIS FORK OF BITCOIN CORE BY LUKEDASHJR.  Type 'yes' to continue! `echo $'\n> '`" yn
```
nothingmuch, line 29: > `THIS FORK OF BITCOIN CORE BY LUKEDASHJR`

AFAIK (i'm not directly participating) this would only be an accurate description of Knots, he's helping this effort, but other people started the bip8 client fork and have also been involved

Tools/toggle_bitcoin_taproot_bip8_lottrue..sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+if [[ "$0" = "$BASH_SOURCE" ]]; then
+    echo "This script must be sourced \". toggle_bitcoin_taproot_bip8_lottrue.sh\"" 
+    exit 1
+fi
+
+set -e
```
nothingmuch, line 7: since this is being sourced, would it make more sense to restore the previous value or to require it to have already been set (e.g. `set -o | grep errexit || exit 1`)

Tools/toggle_bitcoin_taproot_bip8_lottrue..sh
```diff
@@ -0,0 +1,39 @@
...
+
+if [[ "$BTCPAYGEN_ADDITIONAL_FRAGMENTS" =~ "bitcoin-taproot-bip8" ]]; then
+ read -p "The unofficial taproot node release is already active. Type 'official' to change back to the official release `echo $'\n> '`" yn
```
luke-jr, line 10: There is no official release. ST is being deceptively branded "Bitcoin Core" without qualifiers, but it is not following the standard release norms.

Tools/toggle_bitcoin_taproot_bip8_lottrue..sh
luke-jr, line 29: Why ask more than once? The user already typed "unofficial" above.

luke-jr: Indeed, it isn't "by Luke Dashjr" at all. Such wording is dishonest.

Eskyee:
👀

Tools/toggle_bitcoin_taproot_bip8_lottrue..sh
rockstardev: @nothingmuch @luke-jr how about this:

`OK, Are you sure you know what you are doing? This script will swap the Bitcoin Core release with a release provided on https://bitcointaproot.cc signed by LukeDashJr and bitcoinmechanicca@protonmail.com with LOT=TRUE taproot activation. Type 'taproot_lot_true' to switch to this HTTPS://BITCOINTAPROOT.CC FORK OF BITCOIN CORE.`

Tools/toggle_bitcoin_taproot_bip8_lottrue..sh
rockstardev: I think using `bitcoin_core` instead of `official` would be better.

Tools/toggle_bitcoin_taproot_bip8_lottrue..sh
rockstardev: I would consider displaying the first message as informative and then having one confirmation. I can see how `unofficial`+`yes` can be a bit confusing.

NicolasDorier:
@luke-jr do you have a link somewhere to explain to normal people what is the difference between your version and bitcoin core version? Myself I have no idea. If not we can merge as is after the 2 or 3 comments.

Eskyee:
I know of this link @NicolasDorier https://bitcointaproot.cc dropping information on taproot

NicolasDorier:
I agree with the comments, better to replace official by bitcoin_core. If after those comment there is no other objections, merge it.

NicolasDorier:
@Eskyee this is information about taproot, not about how this implementation is different from core.

luke-jr:
>@luke-jr do you have a link somewhere to explain to normal people what is the difference between your version and bitcoin core version? Myself I have no idea.

It's not "my version"; it's the Bitcoin community's Taproot activation release, maintained and published by 3 other community members (I am only helping code/review and advising them on how to do it). The "Bitcoin Core version" is not legitimate, and can be unsafe in some circumstances (though not before November). As for a link, https://bitcointaproot.cc/#faq_mainline

Users should be presented with the BIP8 (bitcointaproot.cc) and BIP9/ST as two equal options (or at least not biased toward the latter because of their deceptive pretending it is more official or legitimate).

NicolasDorier:
User will not be presented with options. It will default to bitcoin core, which minimize chain split risks.
Feel free to also fork this repo if you need otherwise.

luke-jr:
BIP8 minimises chain split risks, not BIP9/ST...

rockstardev (approved):
ACK, @NicolasDorier this is ready to merge

PR #471, opened 2021-04-30
Add new fastsync sig

junderw:


junderw:
@NicolasDorier Here you go!

PR #476, opened 2021-05-04
LND: wumbo fragment

apotdevin:
@rockstardev I know you guys run a fork of LND, would this be the correct param to enable wumbo?


apotdevin:
Closing because I just realized the config is already there by default

PR #485, opened 2021-05-24
Enabling Loop on deployments with legacy LND container

rockstardev:
Give it a try @NicolasDorier


rockstardev:
OK, works for @NicolasDorier.

But in the interest of making this simple, I will wait for 0.13 release of LND. I don't want us to confuse users with multiple docker container names on 0.12.1 version.

PR #489, opened 2021-06-07
Add fragment to enable LND watchtower client

rsafier:


NicolasDorier:
ping @rockstardev should that be just by default or is it better like that?

Also you need to add doc in the README about this fragment.

rsafier:
Added description in README. 
I was surprised to find it wasn't enabled by default. I will be interested in @rockstardev opinion on the direction this should go.

README.md
```diff
@@ -157,6 +157,7 @@ Available `BTCPAYGEN_ADDITIONAL_FRAGMENTS` currently are:
 * [opt-save-storage-xxs](docker-compose-generator/docker-fragments/opt-save-storage-xxs.yml) will keep around 2 weeks of blocks (prune BTC for 5 GB) (lightning not supported)
 * [opt-lnd-autopilot](docker-compose-generator/docker-fragments/opt-lnd-autopilot.yml) will activate auto pilot on LND. (5 channels, 60% of allocation)
 * [opt-lnd-keysend](docker-compose-generator/docker-fragments/opt-lnd-keysend.yml) will activate keysend on LND.
+* [opt-lnd-wtclient](docker-compose-generator/docker-fragments/opt-lnd-wtclient.yml) will activate the watchtower client on LND.
```
Zaxounette, line 160: Active or passive watchtower?
Should specify IMO.

Zaxounette (changes requested):

README.md
Kukks: What does that mean?

Zaxounette: As I understand it, there are two watchtower options available.
Both have the same function, but Passive returns all funds (minus miner fee) whereas Active takes a cut.

It would be useful to know which one (or both?) are activate through the fragment.

rsafier: "Reward watchtowers will be enabled in a subsequent release, though are still undergoing further testing and refinement."

Zaxounette: Okay so what I understand by your comment is that this fragment, by default, is an altruist watchtower.

rsafier: Correct. Tho I misstated the purpose of `wtclient.sweep-fee-rate` in my earlier post, it is simply the fees for the justice transaction the client is sending to the watchtower server. I imagine once the LND team fleshes out how reward watchtowers will operate we will have additional parameters.

rsafier: added ability to define a custom sweep fee rate.

Kukks:
Going to merge this as opt-in for now. We can always make it opt-out or just baked in by default later

PR #490, opened 2021-06-08
Add Tallycoin Connect fragment

dennisreimann:
Integrates the Tallycoin Connect app, so that one can receive Tallycoin payments to the internal BTCPay LND node.

Will clean up links once djbooth007/tallycoin_connect#4 gets merged -- for now I'm publishing the Dockerfiles myself.
Have verified this works on a deployed instance -- inviting more feedback though.

Had to extract the `opt-lnd-grpc` into its own fragment, because otherwise the docker generator would complain.


dennisreimann:
Unfortunately Tallycoin Connect by itself doesn't require authentication, so anyone knowing the URL could access the information provided.

Ideas on how to handle this case? @Kukks @NicolasDorier 

I could add auth requirement in my fork, so we could handle it similarly to how it is done in RTL/ThunderHub.

dennisreimann:
Authentication is now part of Tallycoin Connect and can be enabled by setting the `TALLYCOIN_PASSWD` env variable.

This one is good to go now :)

Kukks (approved):
LGTM!

PR #492, opened 2021-06-22
Fix CentOS7 Compatibility

sumonst21:
fix: Installation error with docker in Centos 7


NicolasDorier:
This break things on non Centos OS `git -c . rev-parse` error is
error: key does not contain a section: .

sumonst21 (commented):
Haven't checked on Centos, this should work better.

sumonst21:
@NicolasDorier Please have a look now `git rev-parse`

NicolasDorier:
Sorry for taking so long

PR #506, opened 2021-08-10
Bump LND to v0.13.1-beta-withloop

rockstardev:
We first need it tested on existing installations before merged, try with testnet server if you have it. Standard commands:

```
git fetch
git checkout feat/lnd-0.13.1-beta-withloop
. btcpay-setup.sh - i
```


rockstardev:
Works on my server™

image

PR #510, opened 2021-08-19
Bump RTL and Clightning-rest

Kukks:
Only for clightning, separate PR for lnd


Kukks:
Just confirmed working

PR #516, opened 2021-08-30
Add Mempool Explorer

Kukks:


dennisreimann:
Why did this never leave draft state? Shall we give it another stab?

dennisreimann:
I think we can pull it off on our own, most of the things @kukks dod look good and I've once also set it up manually. We just need to get to it – docs here. https://github.com/mempool/mempool/tree/master/docker

dennisreimann:
The basics are updated and in place now, need to wait for my txindex to be build so that I can continue here.

dennisreimann:
@softsimon Is there an easy way to run the app under a directory-prefix like `/mempool/` instead of directly in the root? 

The HTML contains `<base href="/">`, which is aso updated via JS depending on the language (e.g. it might get turned into `<base href="/de/">`), which makes the browser load the assets from the root — we'd need a prefix like `/mempool/` to run it in our context, so that we don't have to generate new SSL certs for a subdomain like `mempool.mybtcpay.org`.

Normally we would use nginx `sub_filter` to rewrite paths for our needs, but the JS overrides it anyways.

danielalexiuc:
@kukks @dennisreimann @softsimon I think this is an important feature - I'd like to donate a 0.05 BTC bounty to make it a priority.

dennisreimann:
I have it almost working, fiddling with the details right now.

dennisreimann:
Done now, ready for review :)

Production/nginx.tmpl
```diff
@@ -207,7 +207,44 @@
...
+
+		sub_filter_types *;
+		sub_filter_once off;
+
+		sub_filter 'base href="/'           'base href="/mempool/';
+		sub_filter '/api/v1/'               '/mempool/api/v1/';
+		sub_filter '/resources'             '/mempool/resources';
+
+		proxy_set_header Accept-Encoding "";
+		proxy_pass http://mempool_web:8080/;
```
petzsch, line 228:
```suggestion
```

dupplicate to line 214, leads to nginx crashing!

petzsch (changes requested):
My testing started working after fixing the broken nginx.tmpl

Thanks for this awesome feature. :-)

petzsch:
Viewing Blocks leads to error 404 in the devtools and doesn't show the list of transactions:
https://btcpay.4sats.net/mempool/block/000000000000000000051e3857f4f76691f7897ede16d0e21eee7c2bb60ceb84

image

Production/nginx.tmpl
```diff
@@ -207,7 +207,43 @@
...
+
+		sub_filter_types *;
+		sub_filter_once off;
+
+		sub_filter 'base href="/'           'base href="/mempool/';
+		sub_filter '/api/v1/'               '/mempool/api/v1/';
+		sub_filter '/resources'             '/mempool/resources';
+
+		proxy_set_header Accept-Encoding "";
+
```
dennisreimann, line 228:
```suggestion
```

Production/nginx.tmpl
```diff
@@ -207,7 +207,41 @@
...
+		# Bypass gzip, so that we can apply sub_filter, which does not work with proxy compressed response.
+		# Details: https://www.nginx.com/resources/wiki/modules/substitutions/#directives
+		proxy_set_header Accept-Encoding "";
+
+		sub_filter_types *;
+		sub_filter_once off;
+
+		sub_filter 'base href="/'           'base href="/mempool/';
+		sub_filter '/api/v1/'               '/mempool/api/v1/';
```
dennisreimann, line 224:
```suggestion
		sub_filter '/api/'               '/mempool/api/';
```

dennisreimann:
@petzsch Thanks for the pointer, [0322355](https://github.com/btcpayserver/btcpayserver-docker/pull/516/commits/032235553b8c493bd7180c2b3c1ce355138746b5) contains the fix.

dennisreimann:
## How to test

Go to your BTCPay Server directory and run the following commands:

```bash
# Checkout this pull request
git checkout -b mempool
git pull origin mempool

# Prepare
export BTCPAYGEN_ADDITIONAL_FRAGMENTS="$BTCPAYGEN_ADDITIONAL_FRAGMENTS;opt-add-mempool"

# Install
. ./btcpay-setup.sh -i

# Check
docker logs generated_mempool_api_1
```

ndeet:
Awesome! Unfortunately I couldn't test for now as all my nodes are pruned 😬 
```
Generating /app/Generated/docker-compose.generated.yml
The fragment opt-add-mempool is incompatible with 'pruning'
```

dennisreimann:
@ndeet Yes, the requirement for ElectrumX leads to the incompatibility there. Updated the How To Test comment, thanks for the pointer!

petzsch:
I've noticed a strange redirection bug:

When I want to view a transaction with my browser configured to use the german language by default (i.e. Chrome) and visiting https://btcpay.4sats.net/mempool/tx/9a840e18a6104ebeeee5958027e7377752e51126a1f2f7138bfc1d750dffdd44 - it redirects me to: https://btcpay.4sats.net/mempool/de/

While when the browser is configured to english, it stays on the first link and displays the transaction correctly.

I think the desired behaviour would have been a redirect to https://btcpay.4sats.net/mempool/de/tx/9a840e18a6104ebeeee5958027e7377752e51126a1f2f7138bfc1d750dffdd44

Could be an upstream problem - not sure.

dennisreimann:
@kukks I think we can merge and ship this — the leftover URL issue is minor I think and we can later on fix it.

Kukks:
@danielalexiuc https://d11n.net/donate.html for bounty <3

danielalexiuc:
That's brilliant guys - awesome work, gladly paid. I think a lot of people will benefit from this.

PR #517, opened 2021-08-31
Split Clightning services as individual fragments

Kukks:


pavlenex:
I assume we don't want to pack it all up and deploy by default anymore?

TBH Spark seems abandoned, lightning charge is buggy af...

PR #518, opened 2021-09-01
Provide plugin directory for Clightning

Kukks:
fixes #https://github.com/btcpayserver/btcpayserver/issues/2620


pavlenex:
#2620

PR #523, opened 2021-09-08
Bumping RTL to 0.11.2 due to bug with peering

downet:
Due to bug with adding peers (see https://github.com/Ride-The-Lightning/RTL/issues/245#issuecomment-915538009) bumping version of RTL to 0.11.2


Kukks (approved):
Thanks for the contribution!

PR #548, opened 2021-11-01
Create opt-lnd-reset-onchain.yml

petzsch:
a fragment that allows you to reset the onchain wallet of lnd and trigger a rescan

!!!use with caution!!!


NicolasDorier:
I think this should be added to the doc instead, I think it required non pruned node as well. Leaving it open until we put it in doc.

petzsch:
Gonna close this PR, I've created a issue in the doc repo: https://github.com/btcpayserver/btcpayserver-doc/issues/1106

PR #551, opened 2021-11-03
Added "project-name" param so containers have better names

woutersamaey:
Containers won't have the `generated_` prefix anymore, but `btcpayserver_` instead. The `generated_` prefix comes from the dir name `Generated` where the docker-compose.yml is at and this is default behaviour.

WARNING: This can be tricky and maybe even dangerous if best practices were not followed. Let's discuss.

This change helps to find containers that are part of the BTCPay Server on a server that's running other apps as well.

All I did was add `-p btcpayserver` to all `docker-compose` commands.


Kukks:
I think this would be a breaking change for some scripts.

NicolasDorier:
Yes, while I agree this should have been done from day 1, I am against changing it now as it will breaks quite a lot of things. (documentations and scripts)

PR #552, opened 2021-11-04
C-Lightning REST: Enable access to all RPC commands

danielalexiuc:
The C-Lightning REST project allows access to all additional CLightning RPC commands by default: https://github.com/Ride-The-Lightning/c-lightning-REST#rpc

BTCPayServer-docker currently ships with this option disabled. However many useful RPC commands have not been wrapped by the REST project yet, so it is convenient to enable these by default.

### DIFF
--- a/docker-compose-generator/docker-fragments/bitcoin-clightning.yml
+++ b/docker-compose-generator/docker-fragments/bitcoin-clightning.yml
@@ -61,6 +61,7 @@ services:
       PROTOCOL: "http"
       NODE: clightning_bitcoin:9835
       LIGHTNINGD_READY_FILE: /root/.nbxplorer/btc_fully_synched
+      RPCCOMMANDS: "*"
     links:
       - clightning_bitcoin
     volumes:


NicolasDorier:
why did you close this?

danielalexiuc:
Just doing some more testing on it first, the pull request was a bit early. It seems to work for "no-parameter" rpc calls, but not for calls with parameters.

Will re-open when I have some more information though.

danielalexiuc:
Maybe you can help! Actually yeah the problem seems to be that BTCPayServer won't accept parameters in the json body, which is what is necessary to use this endpoint... any idea where that can be configured?

danielalexiuc:
> Maybe you can help! Actually yeah the problem seems to be that BTCPayServer won't accept parameters in the json body, which is what is necessary to use this endpoint... any idea where that can be configured?

Adding them as request parameters doesn't seem to be an option, it's expecting an array or object, and doesn't seem to parse it as a string.

Kukks:
cln rest is now removed as it was discontinued

PR #558, opened 2021-11-22
Add Monero to opt-save-storage to run pruned node, default Monerod to full node

### LINKED ISSUE #557: Feature Request - Add Monero to opt-save-storage to run pruned node
https://github.com/btcpayserver/btcpayserver-docker/blob/ee2c3f2fcd358bbd8118206098741b4a47aea37f/docker-compose-generator/docker-fragments/opt-save-storage.yml#L5

Monero (monerod) can be run in pruned node, which drops the node size down by about 2/3 with the --prune-blockchain flag.  There aren't really variable sizes that can be used for a pruned node in Monero, so I think opt-save-storage.yml is probably the best place (not 100% sure though). As part of this feature request, Monero should be added to opt-save-storage and BITCOIN_EXTRA_ARGS should be added to the Monerod environment image fragment as it exists in most if not all of the other 'altcoin' fragments.  This will also make Monero env overrides much more in line with the rest of BTCPayServer.

CryptoGrampy:
Resolves #557 

Also to note- this PR would change the default Monerod node fragment behavior from pruned to full (as it should be), and moves that config to opt-save-storage.yml via MONERO_EXTRA_ARGS.  Would appreciate it if someone can review.

### DIFF
--- a/docker-compose-generator/docker-fragments/monero.yml
+++ b/docker-compose-generator/docker-fragments/monero.yml
@@ -5,7 +5,16 @@ services:
     restart: unless-stopped
     container_name: btcpayserver_monerod
     image: btcpayserver/monero:0.17.2.3
-    entrypoint: monerod --rpc-bind-ip=0.0.0.0 --confirm-external-bind --rpc-bind-port=18081 --non-interactive --block-notify="/bin/sh ./scripts/notifier.sh -X GET http://btcpayserver:49392/monerolikedaemoncallback/block?cryptoCode=xmr&hash=%s" --hide-my-port --prune-blockchain --enable-dns-blocklist
+    entrypoint: monerod 
+    environment:
+      MONERO_EXTRA_ARGS: |
+        rpc-bind-ip=0.0.0.0
+        confirm-external-bind=1
+        rpc-bind-port=18081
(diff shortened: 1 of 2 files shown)


CryptoGrampy (commented):
MONEROD -> MONERO

sethforprivacy:
This looks great from my end, thanks for doing this, @CryptoGrampy! 

Pinging @Kukks for review.

Kukks:
Are you sure this wont break existing installs? @sethforprivacy  @CryptoGrampy

sethforprivacy:
> Are you sure this wont break existing installs? @sethforprivacy @CryptoGrampy

If the user was running the previous change (pruning by default) and update to this, their node will remain pruned and continue working properly.

This will only mean that new installs are "full" by-default, instead of pruned by-default.

If @astupidmoose or @jooray could test, though, that would be ideal.

NicolasDorier:
I reverted this commit in https://github.com/btcpayserver/btcpayserver-docker/commit/ec7aa328b79dee5a360a8a50de445f37e7bdd320, it turns out it break monero installs because the docker image doesn't support `MONERO_EXTRA_ARGS`

PR #559, opened 2021-11-24
Added support for Cheat Mode

woutersamaey:


woutersamaey:
@dennisreimann and @NicolasDorier for your review.

dennisreimann (approved):

NicolasDorier:
I think we shouldn't expose it and document it.
What you can do instead is to use a custom fragment.

woutersamaey:
Erm? How does obscurity help anyone?

woutersamaey:
We discussed and agreed this is not a feature for the general public.

NicolasDorier:
https://github.com/btcpayserver/btcpayserver-docker#how-can-i-customize-the-generated-docker-compose-file

PR #560, opened 2021-11-26
Solution to run master branch on docker

woutersamaey:
@NicolasDorier for your review.

### DIFF
--- a/contrib/UsingMasterBranch/README.md
+++ b/contrib/UsingMasterBranch/README.md
@@ -1,0 +1,5 @@
+# Using the `master` branch on your Docker deployment.
+
+## How to use?
+Simply run the script `switch-to-master-and-pull.sh` and everything will be configured for you.
+This script can be put in a cronjob, so your deployment is always on the latest commit from master.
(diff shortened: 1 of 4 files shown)


pavlenex:
Can you ELI5 what is this and why it's useful? Seems very edge-case to be documented in contribute docs.

woutersamaey:
I built this with Nicolas help and he told me to put it here.

This script reconfigures your docker install to pull. build and run the latest code from master.

Nicolas and I will be using it to keep a server running using the latest code for automated testing. In Nicolas' case it will be for future Selenium tests. In my case it will be for Zapier.

pavlenex:
I see. Sorry my bad. I somehow misread that this is doc repository hence the question.

NicolasDorier:
No need of having several fragments. Only one is enough.

The readme should have more explanation about what it is about.

I think the script is not really needed, you can just document the copy of the fragment and the `btcpay-setup.sh` in a code block, in the readme.

Cronjob shouldn't be mentioned. Nothing is related to cron in this, people run it as they see fit.

PR #585, opened 2022-01-25
Upgraded and refactored Traefik

iBobik:
I wrote routing rules only for services I use for now.

Is it ok like this?


docker-compose-generator/docker-fragments/traefik.yml
```diff
@@ -3,23 +3,19 @@ version: "3"
 services:
   traefik:
     restart: unless-stopped
-    image: traefik
+    image: traefik:v2.5
     container_name: traefik
     ports:
       - "${REVERSEPROXY_HTTP_PORT:-80}:80"
       - "${REVERSEPROXY_HTTPS_PORT:-443}:443"
+      - "8080:8080" # Dashboard, enable for debug only
```
Kukks, line 11: Remove or find a way to protect

Kukks (changes requested):
It seems generally good. Thank you so much and sorry for the long delay in reviewing!

* Could you add the additional routes?
* Can we handle the tor integration too somehow?

docker-compose-generator/docker-fragments/traefik.yml
iBobik: It is disabled in traefik.yml by default, you need to uncomment it to enable dashboard.

iBobik:
I do not use Tor, is there somebody who use Tor and has will to experiment with Traefik?

Kukks:
> I do not use Tor, is there somebody who use Tor and has will to experiment with Traefik?

let's do it another time then.

Kukks (approved):

Kukks:
Is there a reason this is still a draft @iBobik

iBobik:
> Is there a reason this is still a draft @iBobik

No, if you are ok with only some services supported. How about to leaving note in a docs like this?
- there are routing rules missing for some services (`traefik.http.routers` labels, PRs welcome)
- TOR support not tested
- when using BTCPAY_ADDITIONAL_HOSTS don't forgot to also add routing rule, see Traefik/README.md

Or maintain all this caveats in just Traefik/README.md and in BTCPay's docs have just link to it?

Kukks:
> > Is there a reason this is still a draft @iBobik
> 
> No, if you are ok with only some services supported. How about to leaving note in a docs like this?
> 
> * there are routing rules missing for some services (`traefik.http.routers` labels, PRs welcome)
> * TOR support not tested
> * when using BTCPAY_ADDITIONAL_HOSTS don't forgot to also add routing rule, see Traefik/README.md
> 
> Or maintain all this caveats in just Traefik/README.md and in BTCPay's docs have just link to it?

ACK

Kukks:
Will merge once that's done. Thanks again for this contrib! :)

NicolasDorier:
@iBobik I needed to revert this on https://github.com/btcpayserver/btcpayserver-docker/commit/9d0edbfa12d9ecb3710cbb011fb6258949af151e because it was bricking some installs with this error during update:

```
ERROR: The Compose file '/root/btcpayserver-docker/Generated/docker-compose.generated.yml' is invalid because:
services.bitcoin_rtl.labels.traefik.enable contains true, which is an invalid type, it should be a string, number, or a null
services.btcpayserver.labels.traefik.enable contains true, which is an invalid type, it should be a string, number, or a null
```

I think changing the boolean to 'true' might work. But I couldn't test.

NicolasDorier:
@iBobik can you just put those labels in `traefik.yml` instead.

Those settings will get then merged by our docker-compose builder in the resulting docker-compose.
Installations not using traeffik won't have the labels.

NicolasDorier:
I attempt doing it on https://github.com/btcpayserver/btcpayserver-docker/commit/f4d7a12488ef5d4c3b83bb069251f536aa34035c I don't know if it works, so please test.

iBobik:
Reason why it is in every service is to have it on eyes for people changing services yamls. It is easy to forgot something they does not use and to break it.

Btw, the same with Nginx env variables.

So I prefer to fix this incompatibility, which we have to do anyway.

For people with bricked instances, do they use some older Docker compose? I have no other idea why it works for most users.

Honza Pobořil
💬Signal

iBobik:
@NicolasDorier Could we talk about this please? I do not agree with your solution.

> Reason why it is in every service is to have it on eyes for people changing services yamls. It is easy to forgot something they does not use and to break it. Btw, the same with Nginx env variables. So I prefer to fix this incompatibility, which we have to do anyway. For people with bricked instances, do they use some older Docker compose? I have no other idea why it works for most users.

Issue #592, opened 2022-02-08
Lightning Terminal pool recovery failure


premghinde:
Steps to reproduce:

1. Add lightning terminal fragment to setup and fund lightning terminal pool with sats
2. In RTL you will see an on chain transaction similar to `poold -- AccountCreation(acct_key=` 
3. Run `btcpay-down.sh`
4. Run `btcpay-up.sh`
5. Log back into lightning terminal and you are now asked to add funds again

Solution:

- Add a volume for the `/root` directory
- Alternative: Add environment variable of `HOME=/data/lit` to compose

premghinde:
For anyone falling foul of this, the command to recover your account inside the lit container is as follows:
`pool --rpcserver=localhost:8443 --macaroonpath /root/.pool/mainnet/pool.macaroon --tlscertpath /root/.lit/tls.cert accounts recover`

PR #602, opened 2022-02-16
Upgrade LiT to v0.6.2 (includes bugfix)

dennisreimann:
Version bump (closes btcpayserver/btcpayserver#3459) and two bugfixes:

- Persist the `LIT_PASSWD` in `.env`
- ac398d569f75a3fb6c9889b1e0f1d75e93725a1c fixes btcpayserver/btcpayserver-docker#601


docker-compose-generator/docker-fragments/opt-add-lightning-terminal.yml
```diff
@@ -4,7 +4,7 @@ services:
     environment:
       BTCPAY_EXTERNALSERVICES: "Lightning Terminal:${BTCPAY_PROTOCOL:-https}://${BTCPAY_HOST}/lit/;"
   lnd_lit:
-    image: "lightninglabs/lightning-terminal:v0.6.1-alpha-path-prefix"
+    image: "lightninglabs/lightning-terminal:v0.6.2-alpha-path-prefix"
```
guggero, line 7: If we're updating this file, shouldn't we also make sure that all of LiT's data is written to the volume correctly? I see there is a volume at `/data/lit`, but LiT isn't configured to use it. Either an ENV var `HOME=/data/lit` should be set (or is this done somewhere else?) or the volume should be mounted to `/root`. See https://github.com/btcpayserver/btcpayserver-docker/issues/592. Currently Pool's data (and I assume Loop, Faraday and Lit's internal data) is lost when the container is re-created.

Production/nginx.tmpl
```diff
@@ -144,15 +144,7 @@
...
-	location ~* ^/lnrpc.Lightning/ {
-    proxy_pass http://lnd_lit:8080;
-  }
-
-  location ~* ^/looprpc.SwapClient/ {
-    proxy_pass http://lnd_lit:8080;
-  }
-
-  location ~* ^/poolrpc.Trader/ {
+  location ~* ^/(ln|loop|pool|lit)rpc\. {
```
guggero, line 147: :tada:

guggero (commented):
Thanks for the update! Changes look good to me, but I think it would be great if we could also address the data/volume issue (see inline comment).

docker-compose-generator/docker-fragments/opt-add-lightning-terminal.yml
dennisreimann: Thanks for the pointer - ok like [cfaed5f](https://github.com/btcpayserver/btcpayserver-docker/pull/602/commits/cfaed5fa23c0ae384b20c5c661c781e7fde6d175)?

guggero: Looks good, yes. Thanks! I'm not able to test this though.

PR #610, opened 2022-02-23
Add JoinMarket Web UI

dennisreimann:
Work in progress state, open TODOS:

- [x] Needs to get merged and released: joinmarket-webui/joinmarket-webui#124
- [ ] Authentication in nginx. The [`.htpasswd` file](https://github.com/btcpayserver/dockerfile-deps/blob/master/JoinMarket/0.9.5/docker-entrypoint.sh#L19) gets created by the joinmarket container, but it isn't integrated into nginx yet


Kukks:
Ideally we also find a way to integrate the cookie auth mechanism instead of the hardcoded password thing too

dennisreimann:
If we can find a general way to integrate this, the web UI teamm would certainly appreciate it, as tehy are still looking for a better way to handle the authentication: joinmarket-webui/joinmarket-webui#120

docker-compose-generator/docker-fragments/opt-add-joinmarket.yml
```diff
@@ -3,10 +3,12 @@ version: "3"
 services:
   joinmarket:
     container_name: joinmarket
-    image: btcpayserver/joinmarket:0.9.4
+    image: btcpayserver/joinmarket:0.9.5
```
apotdevin, line 6: Would be nice to pin this to the image hash as well. Just an additional security measure to avoid overwritten images since afaik they aren't immutable in docker hub

dennisreimann:
Closing because this is outdated.

PR #612, opened 2022-03-01
Update btcpay-setup.sh

iam1412:
git -c for CentOS 7


Kukks (changes requested):
Thank you for the contribution!

I think we've seen this before but it may cause incompatibilities between git versions and different OSes. 

Could you perhaps detect  and use your version only if needed?

iam1412:
Okay

PR #613, opened 2022-03-02
Fixed git -c command for CentOS Linux

iam1412:
Fixed git -C to git -c for CentOS Linux Command

PR #614, opened 2022-03-02
Fixbug: Run install on CentOS

iam1412:
Fixed bug run install on CentOS (Tested on CentOS7,8,9)
- Fixed git -C to git -c
- Fixed confirm remove get-docker.sh
- Fixed start docker and enable service before run "docker run"

### DIFF
--- a/btcpay-setup.sh
+++ b/btcpay-setup.sh
@@ -38,14 +38,28 @@ else
 fi
 
 # Verify we are in right folder. If we are not, let's go in the parent folder of the current docker-compose.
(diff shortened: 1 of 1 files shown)


wakiyamap:
`git -C` and `git -c` are different options.
CentOS 8 and 9 probably work without change.
This is due to an older version of git.
ref. https://github.com/btcpayserver/btcpayserver-docker/pull/613 https://github.com/btcpayserver/btcpayserver-docker/pull/492

I think it's better to consider whether to support CentOS 7.

PR #623, opened 2022-04-05
Fixes "line 30: exec: btcpay-update.sh: not found"

j3th9n:
Fixes "./btcpay-update.sh: line 30: exec: btcpay-update.sh: not found" when running ./btcpay-update.sh


NicolasDorier:
hey sorry for coming back on that so late.
This error isn't a bug, it means that the user didn't run `btcpay-setup.sh -i` or is running `btcpay-update.sh` on the wrong user.

j3th9n:
Thanks for the update and explanation, I'll close this PR.

PR #625, opened 2022-04-18
Add ability to set port when using no proxy

tehelsper:
This helps with automation when deploying BTCPay with an existing bitcoin node


tehelsper:
@NicolasDorier @Kukks Could I get a review / merge on this? It helps automate and improve BTC Pay deployment on myNode.

helpers.sh
```diff
@@ -104,6 +105,8 @@ BTCTRANSMUTER_HOST=$BTCTRANSMUTER_HOST
 CHATWOOT_HOST=$CHATWOOT_HOST
 BTCPAY_CRYPTOS=$BTCPAY_CRYPTOS
 WOOCOMMERCE_HOST=$WOOCOMMERCE_HOST
+REMOTE_BTC_RPC_USERNAME=$REMOTE_BTC_RPC_USERNAME
```
Kukks, line 108: These are not used anywhere?

Kukks (changes requested):
I would split this pr into 2. 

The port stuff is probably fine, if you can make sure it is documented in the Readme as well. 

The other stuff regarding bitcoin node seems unfinished. If you only want additional rpc Auth, you can create a fragment (see Readme) to add it in without needing to do changes here.

helpers.sh
tehelsper: Right, not in the repos. I create a bitcoin.custom.yml file that needs the username and pass. It seems this could help custom YAML files like is recommended in the README.

tehelsper: I didn't want to update the YAML file if the password changes. It was simpler to automate deployment by storing the username / pass as a variable and that way it could be re-used if necessary.

I map it to the existing NBXPLORER_BTCRPCUSER variable in the nbexplorer container. I named it the way I did to indicate the variable wasn't being set, but needs to match the remote btc instance.

```
version: "3"

services:
  btcpayserver:
    environment:
      BTCPAY_CHAINS: "btc"
      BTCPAY_BTCEXPLORERURL: http://nbxplorer:32838/
  nbxplorer:
    environment:
      NBXPLORER_CHAINS: "btc"
      NBXPLORER_BTCRPCURL: http://host.docker.internal:8332/
      NBXPLORER_BTCRPCUSER: "${REMOTE_BTC_RPC_USERNAME}"
      NBXPLORER_BTCRPCPASSWORD: "${REMOTE_BTC_RPC_PASSWORD}"
      NBXPLORER_BTCNODEENDPOINT: host.docker.internal:8333
    volumes:
      - "/mnt/hdd/mynode/bitcoin:/root/.bitcoin"
    extra_hosts:
      - "host.docker.internal:172.17.0.1"
```

tehelsper: @Kukks Thoughts? I can rename if this seems confusing. It seems helpful to be able to set the remote BTC username / pass at setup time.

Kukks: I understand but we can't opinionate the settings based on your own custom fragments, it is already very messy as it is..

One workaround, for now, would be to use the [`env-file` option ](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option)in your fragment and create the env file inside the `Generated` dir

tehelsper: I agree it's currently a bit confusing, but I'd argue this has very little to do with my custom fragment. It's already an env variable, the BTC user/pass is just specific to NBXPLORER and it could be very useful in a larger context which I why I chose that name. I'm just trying to make an existing variable easier to set programmatically.

If I kept the NBXPLORER_BTCRPCUSER name would it be acceptable?

tehelsper: OK, PR updated to remove the RPC variables. Adding a new env vars file works well.

tehelsper: @Kukks Hey, sorry to ping you again, but do you think this is acceptable now? I changed this PR to be port settings only.

Kukks (approved):

PR #641, opened 2022-05-05
Backup and restore process

### LINKED ISSUE #116: Create automated backup

### LINKED ISSUE #249: Back up account, config files and keep all settings
Hi, I have installed BTCPayServer using the docker-compose manual guide on Linux Ubuntu 18.04.

Let's say I now want to back up the whole server and move it to another computer.
What folders would I need to back up so I would keep my account and all the settings?

/var/lib/docker/volumes (which is quite large)
/etc/docker
$BTCPAY_ENV_FILE
$BTCPAY_DOCKER_COMPOSE
/etc/profile.d/btcpay-env.sh

Anything else I am missing?

### LINKED ISSUE #496: Backup script is not working and restoration process is not documented
**Describe the bug**
- backup.sh is not fully working
- restoration process is not document

**To Reproduce the bug**
1. execute backup.sh
2. postgresql dump will fail
3. no procedure or script to do restoration

**Expected behavior**

1. create backup
2. intall a new server
3. restore the backup

**Your BTCPay Environment (please complete the following information):**
- BTCPay Server Version:  1.1.2.0
- Deployment Method: docker
- Browser: chrome and safari, but they are not related with the problem

**Logs (if applicable)**
no

**Setup Parameters**
default parameters with just the domain changed

**Additional context**
Is clear for me that backup script stop the btcpayserver service, but not the docker, compress some directories with some exceptions, do postgresql dump and compress together with the backup. Based on that, I suppose that restorarion process will be the same: stop btcpayserver, extract the files inside volumes directory and start the service. I have some questions that I think will help a lot the btcpay community:

1.  backup.sh is failing to dumo the postgresql database. (bug)
2. if there is a backup.sh, why do not offer restore.sh?
3. supposing that backup.sh can dump the postgresql database; if I extract the files and recover the postgresql, is it enough?
4. as it is using docker, why I can not just stop btcpayserver service and docker service and move entire /var/lib/docker to the new server? Here I am more concerned about migration best practices. I understad 
5. is really necessary to dump posgresql database if in the new server i have the same postgresql version? because docker volumes will contain all postgresql data.
6. Can someone enumerate safe steps to dump the data and recover in the new server? For me is clear when we are using docker volumes, you just have to be concerned about the volumes. But I tested this and did not work.
7. Do I need to wait the btc sync to do the migration (export and import the backup)? I read some users saying that in btcpayserver git and i do not see a reason to do that.
8. Can someone the safe way to dump and import postgresql? I understand this is just a docker/postgresql, but at the same time the official backup.sh is not working.
9. Which is the correct way to change the btcpayserver image? i am doing systemctl stop btcpayserver, ./btcpay-update.sh and then systemctl start btcpayserver. I tried to use ./btcpay-restart.sh, but I started to see nginx 502 errors.  This procedure must be clear to backups be realiable.

I read a lot of issues and related sites with btcpayserver gits and I do not think backup, migration and restoration process are clear for everyone. I am here to test everything and give good feedbacks. I did not have success yet with migrations, but I am sure the most of people are not just relying in cloud snapshots.

### LINKED ISSUE #628: backups dump postgresql database then fail to include it
I looked at `backup.sh` and it's pretty bizarre. First it does this:

```
# dump database
echo "Dumping database …"
btcpay_dump_db $dumpname
```

which is really just

```
docker exec $(docker ps -a -q -f "name=postgres_1") pg_dumpall -c -U postgres >/var/lib/docker/volumes/backup_datadir/_data/postgres.sql
```

Then it stops the server (which means the backup could miss transactions in the database after the dump), then:

```
    tar --exclude="$backup_dir/*" --exclude="$volumes_dir/generated_bitcoin_datadir/*" --exclude="$volumes_dir/generated_litecoin_datadir/*" --exclude="$volumes_dir/**/logs/*" -cvzf $backup_path $dbdump_path $volumes_dir
```

If you look closely, you'll notice it excludes backups from the tar file (`--exclude="$backup_dir/*"`). The `$dbdump_path` path is passed in there, presumably as an attempt to override that, but it doesn't work: there's no SQL dump in the backup file.

### LINKED ISSUE #629: backup procedure has no restore procedure
on top of being somewhat inconsistent (#628), the backup procedure doesn't have an associated restore procedure, which is dearly missing. i was told on mattermost this was being "worked on" but "in early stages".

Here's the procedure I wrote for our internal docs:

> A manual backup/restore procedure might look like this:
> 
>     systemctl stop btcpayserver
>     tar cfz backup.tgz /var/lib/docker/volumes/
>     systemctl start btcpayserver
> 
> A restore, then, would look like this:
> 
>     systemctl stop btcpayserver
>     mv /var/lib/docker/volumes/ /var/lib/docker/volumes.old # optional
>     tar -C / -x -f -z backup.tgz
>     systemctl start btcpayserver
> 
> The `mv` step should be turned into a `rm -rf
> /var/lib/docker/volumes/` command if we are likely to run out of disk
> space on restore and we're confident in the backup's integrity.
> 
> If you're worried about the backup clobbering other files on restore
> (for example you're not sure about the backup source or file
> structure), this should restore only `volumes/` in the
> `/var/lib/docker` directory:
> 
>     systemctl stop btcpayserver
>     mv /var/lib/docker/volumes/ /var/lib/docker/volumes.old # optional
>     tar -C /var/lib/docker/ -x -f backup.tar.gz -z --strip-components=3
>     systemctl start btcpayserver
>

dennisreimann:
This is a fresh approach to backup and restore a BTCPay Server instance. The existing `backup.sh` script will stay (mostly) untouched and we are providing two new scripts:

- `btcpay-backup.sh`, which produces a `backup.tar.gz` file
- `btcpay-restore.sh`, which takes the path to the backup file and recreates the instance

We won't nail every aspect right away, so the goal of this PR is to provide a solid base which can then be extended in follow-up PRs. Right now, the script dumps the database to a `postgres.sql` file and backs it up together with the `volumes/generated_*` files (except the Bitcoin/Litecoin blockchain and Postgres directories).

## TODO

- [x] Let the backup script crash if something goes wrong instead of continuing like everything is fine
- [x] Use "docker volume inspect" to find the actual path of the volumes to back up instead of assuming they are in the standard dir. 
- [x]  (Optionally) [encrypt the backup](https://linuxconfig.org/how-to-create-compressed-encrypted-archives-with-tar-and-gpg) file for safe offsite storage using a passphrase and strong encryption (GPG / RSA)
- [x] Provide decent error messages when things go wrong.
- [x] Restore procedure
  - [x] Database should only be accessed by the restore process
  - [x] Handle existing data/folder (remove?) and recreate volumes
- [x] Document backup and restore
  - Tracked in btcpayserver/btcpayserver-doc#1116
  - Worked on in btcpayserver/btcpayserver-doc#1121

## Further ideas

- Also back up the `btcpayserver-docker` directory. 
- Also contain the project & tooling (so not just the data). This is important because an old backup may not work on newer code.
- Find a better or more appropriate way to backup the LN node static channel data
- The current `backup.sh` script does Dropbox uploads and such. Those things could be enabled here as well, but with a more modular approach, for instance by providing wrapper scripts for the new `btcpay-backup.sh`.

## Related issues

- Closes #628
- Closes #629 
- Closes #496
- Closes #249
- Closes #116

### DIFF
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 
 # Please be aware of these important issues:
 #
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Please be aware of these important issues:
+#
+# - Old channel state is toxic and you can loose all your funds, if you or someone
+#   else closes a channel based on the backup with old state - and the state changes
+#   often! If you publish an old state (say from yesterday's backup) on chain, you
+#   WILL LOSE ALL YOUR FUNDS IN A CHANNEL, because the counterparty will publish a
+#   revocation key!
+
+if [ "$(id -u)" != "0" ]; then
+  echo "This script must be run as root."
+  echo "Use the command 'sudo su -' (include the trailing hypen) and try again"
+  exit 1
+fi
+
+# preparation
+docker_dir=/var/lib/docker
+dbdump_name=postgres.sql
+btcpay_dir="$BTCPAY_BASE_DIRECTORY/btcpayserver-docker"
+backup_dir="$docker_dir/volumes/backup_datadir/_data"
+dbdump_path="$docker_dir/$dbdump_name"
+backup_path="$backup_dir/backup.tar.gz"
+
+# ensure backup dir exists
+if [ ! -d "$backup_dir" ]; then
+  docker volume create backup_datadir
+fi
+
+cd $btcpay_dir
+. helpers.sh
+
+echo "Stopping BTCPay Server …"
+btcpay_down
+
+echo "Dumping database …"
+btcpay_dump_db $dbdump_path
+
+echo "Backing up files …"
+cd $docker_dir
+tar \
+  --exclude="volumes/backup_datadir" \
+  --exclude="volumes/generated_postgres_datadir" \
+  --exclude="volumes/generated_bitcoin_datadir" \
+  --exclude="volumes/generated_litecoin_datadir" \
+  --exclude="**/logs/*" \
+  -cvzf $backup_path $dbdump_name volumes/generated_*
+cd -
+
+echo "Restarting BTCPay Server …"
+btcpay_up
+
+echo "Cleaning up …"
+rm $dbdump_path
+
+echo "Backup done."
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -1,5 +1,7 @@
 #!/bin/bash -e
 
+set -o pipefail -o errexit
+
 # Please be aware of these important issues:
 #
 # - Old channel state is toxic and you can loose all your funds, if you or someone
@@ -43,7 +45,7 @@ if [ -z "$dbcontainer" ]; then
 fi
 
 echo "Dumping database …"
-btcpay_dump_db $dbdump_path
+docker exec $dbcontainer pg_dumpall -c -U postgres | gzip > $dbdump_path
 
 echo "Stopping BTCPay Server …"
 btcpay_down
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -30,11 +30,23 @@ fi
 cd $btcpay_dir
 . helpers.sh
 
-echo "Stopping BTCPay Server …"
-btcpay_down
+dbcontainer=$(docker ps -a -q -f "name=postgres_1")
+if [ -z "$dbcontainer" ]; then
+  echo "Database container is not up and running. Starting BTCPay Server …"
+  btcpay_up
+
+  dbcontainer=$(docker ps -a -q -f "name=postgres_1")
+  if [ -z "$dbcontainer" ]; then
+    echo "Database container could not be started or found."
+    exit 1
+  fi
+fi
 
 echo "Dumping database …"
 btcpay_dump_db $dbdump_path
+
+echo "Stopping BTCPay Server …"
+btcpay_down
 
 echo "Backing up files …"
 cd $docker_dir
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -16,7 +16,7 @@ fi
 
 # preparation
 docker_dir=$(docker volume inspect generated_btcpay_datadir --format="{{.Mountpoint}}" | sed -e "s%/volumes/.*%%g")
-dbdump_name=postgres.sql
+dbdump_name=postgres.sql.gz
 btcpay_dir="$BTCPAY_BASE_DIRECTORY/btcpayserver-docker"
 backup_dir="$docker_dir/volumes/backup_datadir/_data"
 dbdump_path="$docker_dir/$dbdump_name"
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -11,8 +11,8 @@ set -o pipefail -o errexit
 #   revocation key!
 
 if [ "$(id -u)" != "0" ]; then
-  echo "This script must be run as root."
-  echo "Use the command 'sudo su -' (include the trailing hypen) and try again"
+  printf "\n🚨 This script must be run as root.\n"
+  printf "➡️  Use the command 'sudo su -' (include the trailing hypen) and try again.\n\n"
   exit 1
 fi
 
@@ -34,37 +34,70 @@ cd $btcpay_dir
 
 dbcontainer=$(docker ps -a -q -f "name=postgres_1")
 if [ -z "$dbcontainer" ]; then
-  echo "Database container is not up and running. Starting BTCPay Server …"
+  printf "\n"
+  echo "ℹ️  Database container is not up and running. Starting BTCPay Server …"
   btcpay_up
 
+  printf "\n"
   dbcontainer=$(docker ps -a -q -f "name=postgres_1")
   if [ -z "$dbcontainer" ]; then
-    echo "Database container could not be started or found."
+    echo "🚨 Database container could not be started or found."
     exit 1
   fi
 fi
 
-echo "Dumping database …"
-docker exec $dbcontainer pg_dumpall -c -U postgres | gzip > $dbdump_path
+printf "\n"
+echo "ℹ️  Dumping database …"
+{
+  docker exec $dbcontainer pg_dumpall -c -U postgres | gzip > $dbdump_path
+  echo "✅ Database dump done."
+} || {
+  echo "🚨 Dumping failed. Please check the error message above."
+  exit 1
+}
 
-echo "Stopping BTCPay Server …"
+printf "\nℹ️  Stopping BTCPay Server …\n\n"
 btcpay_down
 
-echo "Backing up files …"
+printf "\n"
 cd $docker_dir
-tar \
-  --exclude="volumes/backup_datadir" \
-  --exclude="volumes/generated_postgres_datadir" \
-  --exclude="volumes/generated_bitcoin_datadir" \
-  --exclude="volumes/generated_litecoin_datadir" \
-  --exclude="**/logs/*" \
-  -cvzf $backup_path $dbdump_name volumes/generated_*
-cd -
+echo "ℹ️  Archiving files in $(pwd)…"
 
-echo "Restarting BTCPay Server …"
+{
+  tar \
+    --exclude="volumes/backup_datadir" \
+    --exclude="volumes/generated_bitcoin_datadir" \
+    --exclude="volumes/generated_litecoin_datadir" \
+    --exclude="volumes/generated_postgres_datadir" \
+    --exclude="volumes/generated_clightning_bitcoin_datadir/_data/lightning-rpc" \
+    --exclude="**/logs/*" \
+    -cvzf $backup_path $dbdump_name volumes/generated_*
+  echo "✅ Archive done."
+
+  if [ ! -z "$BTCPAY_BACKUP_PASSPHRASE" ]; then
+    printf "\n"
+    echo "🔐 BTCPAY_BACKUP_PASSPHRASE is set, the backup will be encrypted."
+    {
+      gpg -o "$backup_path.gpg" --batch --yes -c --passphrase "$BTCPAY_BACKUP_PASSPHRASE" $backup_path
+      rm $backup_path
+      backup_path="$backup_path.gpg"
+      echo "✅ Encryption done."
+    } || {
+      echo "🚨  Encrypting failed. Please check the error message above."
+      # do not exit, we need to restart BTCPay Server
+    }
+  fi
+} || {
+  echo "🚨  Archiving failed. Please check the error message above."
+  # do not exit, we need to restart BTCPay Server
+}
+
+printf "\nℹ️  Restarting BTCPay Server …\n\n"
+
+cd -
 btcpay_up
 
-echo "Cleaning up …"
+printf "\nℹ️  Cleaning up …\n\n"
 rm $dbdump_path
 
-echo "Backup done."
+printf "✅ Backup done => $backup_path\n\n"
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -26,7 +26,7 @@ backup_path="$backup_dir/backup.tar.gz"
 
 # ensure backup dir exists
 if [ ! -d "$backup_dir" ]; then
-  docker volume create backup_datadir
+  mkdir -p $backup_dir
 fi
 
 cd $btcpay_dir
@@ -36,7 +36,8 @@ dbcontainer=$(docker ps -a -q -f "name=postgres_1")
 if [ -z "$dbcontainer" ]; then
   printf "\n"
   echo "ℹ️  Database container is not up and running. Starting BTCPay Server …"
-  btcpay_up
+  docker volume create generated_postgres_datadir
+  docker-compose -f $BTCPAY_DOCKER_COMPOSE up -d postgres
 
   printf "\n"
   dbcontainer=$(docker ps -a -q -f "name=postgres_1")
@@ -84,17 +85,22 @@ echo "ℹ️  Archiving files in $(pwd)…"
       echo "✅ Encryption done."
     } || {
       echo "🚨  Encrypting failed. Please check the error message above."
-      # do not exit, we need to restart BTCPay Server
+      printf "\nℹ️  Restarting BTCPay Server …\n\n"
+      cd $btcpay_dir
+      btcpay_up
+      exit 1
     }
   fi
 } || {
   echo "🚨  Archiving failed. Please check the error message above."
-  # do not exit, we need to restart BTCPay Server
+  printf "\nℹ️  Restarting BTCPay Server …\n\n"
+  cd $btcpay_dir
+  btcpay_up
+  exit 1
 }
 
 printf "\nℹ️  Restarting BTCPay Server …\n\n"
-
-cd -
+cd $btcpay_dir
 btcpay_up
 
 printf "\nℹ️  Cleaning up …\n\n"
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -15,7 +15,7 @@ if [ "$(id -u)" != "0" ]; then
 fi
 
 # preparation
-docker_dir=/var/lib/docker
+docker_dir=$(docker volume inspect generated_btcpay_datadir --format="{{.Mountpoint}}" | sed -e "s%/volumes/.*%%g")
 dbdump_name=postgres.sql
 btcpay_dir="$BTCPAY_BASE_DIRECTORY/btcpayserver-docker"
 backup_dir="$docker_dir/volumes/backup_datadir/_data"
--- a/btcpay-restore.sh
+++ b/btcpay-restore.sh
@@ -9,12 +9,12 @@ fi
 backup_file=$1
 if [ -z "$backup_file" ]; then
   echo "Usage: btcpay-restore.sh /path/to/backup.tar.gz"
-  exit
+  exit 1
 fi
 
 if [ ! -f "$backup_file" ]; then
   echo "$backup_file does not exist."
-  exit
+  exit 1
 fi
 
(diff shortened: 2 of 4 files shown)


woutersamaey:
@dennisreimann the first point "Let the backup script crash if something goes wrong instead of continuing like everything is fine" can be acchieved easily by adding `set -e` at the top of the bash scripts. If the command return an exit code that is non-0 it should be fine. Careful though that we are not doing any strange things that may discard the exit codes...

btcpay-backup.sh
```diff
@@ -0,0 +1,56 @@
...
+#   revocation key!
+
+if [ "$(id -u)" != "0" ]; then
+  echo "This script must be run as root."
+  echo "Use the command 'sudo su -' (include the trailing hypen) and try again"
+  exit 1
+fi
+
+# preparation
+docker_dir=/var/lib/docker
```
woutersamaey, line 18: Can we avoid hard coding this?

btcpay-restore.sh
```diff
@@ -0,0 +1,24 @@
...
+if [ "$(id -u)" != "0" ]; then
+  echo "This script must be run as root."
+  echo "Use the command 'sudo su -' (include the trailing hypen) and try again"
+  exit 1
+fi
+
+backup_file=$1
+if [ -z "$backup_file" ]; then
+  echo "Usage: btcpay-restore.sh /path/to/backup.tar.gz"
+  exit
```
woutersamaey, line 12: use `exit 1` since it didn't really restore anything.

btcpay-restore.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+if [ ! -f "$backup_file" ]; then
+  echo "$backup_file does not exist."
+  exit
```
woutersamaey, line 17: use `exit 1` since it didn't really restore anything.

btcpay-restore.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+volumes_dir=/var/lib/docker/volumes
```
woutersamaey, line 20: Best not to hardcode

helpers.sh
```diff
@@ -174,11 +174,7 @@ btcpay_restart() {
...
     pushd . > /dev/null
     cd "$(dirname "$BTCPAY_ENV_FILE")"
-    backup_dir="/var/lib/docker/volumes/backup_datadir/_data"
-    if [ ! -d "$backup_dir" ]; then
-        docker volume create backup_datadir
-    fi
-    local filename=${1:-"postgres-$(date "+%Y%m%d-%H%M%S").sql"}
-    docker exec $(docker ps -a -q -f "name=postgres_1") pg_dumpall -c -U postgres > "$backup_dir/$filename"
+    local file_path=${1:-"postgres-$(date "+%Y%m%d-%H%M%S").sql"}
+    docker exec $(docker ps -a -q -f "name=postgres_1") pg_dumpall -c -U postgres > "$file_path"
```
woutersamaey, line 178: Can we dump directly to a `.sql.gz` file? If the DB is huge, you could run out of disk space during the backup process. Usually it's as easy as adding a pipe to `gzip --best` or `gzip --fast` and piping again when importing.
Regular sql files can be huge so I usually avoid them at all costs and work directly with a compressed file in 1 go.
Might be nitpicking though as BTCPay DBs are not that big most of the time.

woutersamaey (commented):
Some comments.
Have not run it though.

woutersamaey:
IMHO we can drop the last "Further ideas" point of uploading do dropbox etc.
Moving the backup to an external location is not really part of this script's duty and a sysadmin can very easily add his own line of code to copy/move the file wherever he likes.

woutersamaey:
Question: can these scripts be run from a crontab? Reason I'm asking is because of environment vars that may be or may not be present, i.e. if the crontab runs as "root" for example. We must test this to be sure. Also, cronjobs run non-interactively so any docker commands must run without the `-it` parameter. Not sure if this is done here, but making a note regardless. It's a common mistake.

helpers.sh
ndeet: Agree would defenitely do it, does not cost much but is a saver if the db gets big as the compression is very good even in standard mode. 20GB mysql db dump is ~ 1GB gzipped.

gzip on dump:
`dumpcommand | gzip > file_path`

So for above should be something like this (untested, yolo):
```
local file_path=${1:-"postgres-$(date "+%Y%m%d-%H%M%S").sql.gz"}
docker exec $(docker ps -a -q -f "name=postgres_1") pg_dumpall -c -U postgres | gzip > "$file_path"
```

on restore gunzip and import:
`gunzip < path_to.sql.gz | db_import command`

btcpay-restore.sh
ndeet: in the PR draft it is mentioned as todo :)

helpers.sh
woutersamaey: One thing to be mindful of here is the exit code. If the dump command fails, you will be creating a empty `.sql.gz` file and the backup script won't know something failed, so this can be a serious issue.
The file would contain the error message and be very small.

I can think of 2 solutions:
- Checking the file size. Should be more than a few kB (ugly check)
- Maybe we can pipe only the stdout and leave stderr alone? Not sure if possible and what it would look like. Also depends on the dump command if it respects stdout and stderr.

btcpay-backup.sh
dennisreimann: Done in [030faa7](https://github.com/btcpayserver/btcpayserver-docker/pull/641/commits/030faa7717746ca7210a30be67df68916db9a77a).

btcpay-restore.sh
dennisreimann: Done in [0444ff3](https://github.com/btcpayserver/btcpayserver-docker/pull/641/commits/0444ff3e69c703dc4e1e3c767135d337c15af66c).

btcpay-restore.sh
dennisreimann: Done in [0444ff3](https://github.com/btcpayserver/btcpayserver-docker/pull/641/commits/0444ff3e69c703dc4e1e3c767135d337c15af66c).

helpers.sh
dennisreimann: Done in [05d96f4](https://github.com/btcpayserver/btcpayserver-docker/pull/641/commits/05d96f4887a91781cb61c894b7b4e8bef4a17a8b).

woutersamaey: Did you check what happens if the dump fails? Does the backup crash or do you get a dump file containing an error (and not a working backup)?

dennisreimann:
@woutersamaey yes, I'm running the backup script via crontab. As long as the correct `PATH` is set it works, because the script sources the `.env` file itself.

Example:

```
SHELL=/bin/bash
PATH=/bin:/usr/sbin:/usr/bin:/usr/local/bin
15 4 * * * /root/btcpayserver-docker/btcpay-backup.sh >/dev/null 2>&1
```

helpers.sh
dennisreimann: I've now added [fa503a5](https://github.com/btcpayserver/btcpayserver-docker/pull/641/commits/fa503a50c4af1b50d571748c8f9e2a0ab3863910), which ensures that. To my knowledge this is the only way to achieve this, at least if we want to circumvent saving the unzipped version separately. ([details](https://unix.stackexchange.com/questions/385089/how-to-gzip-only-if-mysqldump-does-not-error/451044))

dennisreimann:
Added encryption and decryption with the optional `BTCPAY_BACKUP_PASSPHRASE`. 
Also improved the log messages and error handling.

Sample output of the scripts …

## Backup

```
ℹ️  Dumping database …
✅ Database dump done.

ℹ️  Stopping BTCPay Server …

Stopping letsencrypt-nginx-proxy-companion     ... done
Stopping nginx-gen                             ... done
Stopping generated_btcpayserver_1              ... done
Stopping generated_clightning_bitcoin_charge_1 ... done
Stopping generated_clightning_bitcoin_spark_1  ... done
Stopping generated_clightning_bitcoin_rest_1   ... done
Stopping btcpayserver_clightning_bitcoin       ... done
Stopping btcpayserver_bitcoind                 ... done
Stopping tor-gen                               ... done
Stopping generated_nbxplorer_1                 ... done
Stopping generated_postgres_1                  ... done
Stopping tor                                   ... done
Stopping generated_bitcoin_rtl_1               ... done
Stopping nginx                                 ... done
Removing letsencrypt-nginx-proxy-companion     ... done
Removing nginx-gen                             ... done
Removing generated_btcpayserver_1              ... done
Removing generated_clightning_bitcoin_charge_1 ... done
Removing generated_clightning_bitcoin_spark_1  ... done
Removing generated_clightning_bitcoin_rest_1   ... done
Removing btcpayserver_clightning_bitcoin       ... done
Removing btcpayserver_bitcoind                 ... done
Removing tor-gen                               ... done
Removing generated_nbxplorer_1                 ... done
Removing generated_postgres_1                  ... done
Removing tor                                   ... done
Removing generated_bitcoin_rtl_1               ... done
Removing nginx                                 ... done
Removing network generated_default

ℹ️  Archiving files in /var/lib/docker…
postgres.sql.gz
volumes/generated_bitcoin_wallet_datadir/
volumes/generated_bitcoin_wallet_datadir/_data/
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/.walletlock
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/wallet.dat
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/db.log
volumes/generated_btcpay_datadir/
volumes/generated_btcpay_datadir/_data/
volumes/generated_btcpay_datadir/_data/FastSynced
volumes/generated_btcpay_datadir/_data/LocalStorage/
volumes/generated_btcpay_datadir/_data/LocalStorage/tmp/
volumes/generated_btcpay_datadir/_data/btcpay20220512.log
volumes/generated_btcpay_datadir/_data/tmp/
volumes/generated_btcpay_datadir/_data/host_id_rsa
volumes/generated_btcpay_datadir/_data/Main/
volumes/generated_btcpay_datadir/_data/Main/settings.config
volumes/generated_btcpay_datadir/_data/host_id_rsa.pub
volumes/generated_btcpay_datadir/_data/host_authorized_keys
volumes/generated_btcpay_datadir/_data/key-47ec2b82-ed01-45ef-8a63-390470cc10e5.xml
[...]
volumes/generated_tor_torrcdir/
volumes/generated_tor_torrcdir/_data/
volumes/generated_tor_torrcdir/_data/torrc
volumes/generated_tor_torrcdir/_data/torrc-2
✅ Archive done.

🔐 BTCPAY_BACKUP_PASSPHRASE is set, the backup will be encrypted.
✅ Encryption done.

ℹ️  Restarting BTCPay Server …

/root/btcpayserver-docker
Creating network "generated_default" with the default driver
Creating nginx                   ... done
Creating generated_postgres_1    ... done
Creating generated_bitcoin_rtl_1 ... done
Creating tor                     ... done
Creating btcpayserver_bitcoind   ... done
Creating tor-gen                         ... done
Creating generated_nbxplorer_1           ... done
Creating btcpayserver_clightning_bitcoin ... done
Creating generated_clightning_bitcoin_charge_1 ... done
Creating generated_clightning_bitcoin_spark_1  ... done
Creating generated_btcpayserver_1              ... done
Creating generated_clightning_bitcoin_rest_1   ... done
Creating nginx-gen                             ... done
Creating letsencrypt-nginx-proxy-companion     ... done

ℹ️  Cleaning up …

✅ Backup done => /var/lib/docker/volumes/backup_datadir/_data/backup.tar.gz.gpg

```

## Restore

```
ℹ️  Cleaning restore directory /var/lib/docker/volumes/backup_datadir/_data/restore …

🔐 Decrypting backup file …
gpg: AES256.CFB encrypted data
gpg: encrypted with 1 passphrase
✅ Decryption done.

ℹ️  Extracting files in /var/lib/docker/volumes/backup_datadir/_data/restore…
postgres.sql.gz
volumes/generated_bitcoin_wallet_datadir/
volumes/generated_bitcoin_wallet_datadir/_data/
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/.walletlock
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/wallet.dat
volumes/generated_bitcoin_wallet_datadir/_data/mainnet/db.log
volumes/generated_btcpay_datadir/
volumes/generated_btcpay_datadir/_data/
volumes/generated_btcpay_datadir/_data/FastSynced
volumes/generated_btcpay_datadir/_data/LocalStorage/
volumes/generated_btcpay_datadir/_data/LocalStorage/tmp/
volumes/generated_btcpay_datadir/_data/btcpay20220512.log
volumes/generated_btcpay_datadir/_data/tmp/
volumes/generated_btcpay_datadir/_data/host_id_rsa
volumes/generated_btcpay_datadir/_data/Main/
volumes/generated_btcpay_datadir/_data/Main/settings.config
volumes/generated_btcpay_datadir/_data/host_id_rsa.pub
volumes/generated_btcpay_datadir/_data/host_authorized_keys
volumes/generated_btcpay_datadir/_data/key-47ec2b82-ed01-45ef-8a63-390470cc10e5.xml
[...]
volumes/generated_tor_torrcdir/
volumes/generated_tor_torrcdir/_data/
volumes/generated_tor_torrcdir/_data/torrc
volumes/generated_tor_torrcdir/_data/torrc-2

ℹ️  Restoring database …--
-- PostgreSQL database cluster dump
--
SET default_transaction_read_only = off;
SET
SET client_encoding = 'UTF8';
SET
SET standard_conforming_strings = on;
SET
--
-- Drop databases (except postgres and template1)
--
DROP DATABASE btcpayservermainnet;
DROP DATABASE nbxplorermainnet;
ERROR:  database "btcpayservermainnet" is being accessed by other users
DETAIL:  There is 1 other session using the database.

[...]

--
-- PostgreSQL database dump complete
--
--
-- PostgreSQL database cluster dump complete
--

ℹ️  Cleaning up …

✅ Restore done
```

woutersamaey:
Looking awesome @dennisreimann !

Are we confident about using this with a LN node too?

dennisreimann:
As long as one does not publish an outdated/inaccurate channel state it should be fine. That's not something we can prevent I think.

woutersamaey:
You mean to say it is fine as long as you don't suffer from bad timing? i.e. a channel was just opened after you made the backup and everything went to hell? So on a low volume system you'd be fine 99.9% of the time? Or am I misinterpreting this?

dennisreimann:
What I'm saying is that there are use cases that are beyond the scope of these scripts. 

For Lightning they will work in the migration case, where you do a clean shutdown and restore the backup on the new instance. It has likely chances to fail in the disaster recovery scenario, where you maybe backup once per night and need to restore that backup. 

The Lightning static channel backup for instance should be watched by a script and copied over to a remote server, to ensure you always have the latest state available. 

We can add a separate script for that use case, but it depends on the implementation and there isn't an easy solution, especially as the user has to provide more input for the remote connection.

woutersamaey:
Are you saying this script is for migration only and not for nightly backups then?

For users who do not use LN, they can use it as a reliable backup process, no?

Kukks:
> Are you saying this script is for migration only and not for nightly backups then?
> 
> For users who do not use LN, they can use it as a reliable backup process, no?

You cannot use this as a backup process for LN.

dennisreimann:
> Are you saying this script is for migration only and not for nightly backups then?

It is also for nightbly backups, but for LN you'll need an extra/additional backup mechanism that works as I explained above: Whenever the channel state changes, a snapshot needs to be taken and at best get stored remotely. 

The different implementations added or are working on support for putting the channel state into a Postgres DB. This might simplify that in the long run, but we aren't there yet.

> For users who do not use LN, they can use it as a reliable backup process, no?

Once done, this should be a reliable process for everyone, just with the caveat explained above.

woutersamaey:
From what I understand, LND is already supporting PostgreSQL on new installations. Does this mean new installs are already OK, or is more work needed in BTCPay to make the connection?

I understand Core Lightning and Eclair are different.

dennisreimann:
There's more work needed. As said, I think this should be handled by different scripts as that topic has a differnt set of complexity on its own.

petzsch:
> IMHO we can drop the last "Further ideas" point of uploading do dropbox etc. Moving the backup to an external location is not really part of this script's duty and a sysadmin can very easily add his own line of code to copy/move the file wherever he likes.

While I agree that providing the entire solution of remote backups in the script or the docs is out of scope, I think a good middle ground would be to provide a link to a good solution to i.e. duply or duplicity (which by the way also handle backup encryption with gpg)

NicolasDorier:
@dennisreimann note that you are not backing up wallets in bitcoind/litecoind there might have wallets with real money there if using eclair for example.

dennisreimann:
@NicolasDorier we should document that in the Considerations section of the docs.

btcpay-restore.sh
```diff
@@ -0,0 +1,138 @@
...
+  btcpay_up
+  exit 1
+}
+
+printf "\nℹ️  Restarting BTCPay Server …\n\n"
+cd $btcpay_dir
+btcpay_up
+
+printf "\nℹ️  Cleaning up …\n\n"
+# rm -rf $restore_dir
```
ndeet, line 135: As this and the following line is commented nothing gets deleted, so maybe also comment the above "Cleaning up" output if there is nothing removed and instead tell them to remove manually or do I miss something?

ndeet:
Awesome work!

When reading the docs I wondered if we should we add some note to remind users that they need to update their dns records to the new IP if they move to a different host (e.g. migration to different server)? 
Also not sure what will happen to the .onion address, will it stay the same or change?

btcpay-restore.sh
dennisreimann: My bad, this line should not have been commented out.

dennisreimann: Fixed in #655.

PR #648, opened 2022-05-12
Exclude all "generated_" dirs from backup

bolatovumar:
Addresses issue raised in https://github.com/btcpayserver/btcpayserver/discussions/3201

Looking at the [script used by LunaNode](https://github.com/LunaNode/launchbtcpay/blob/master/run.sh) there can be many more "generated_*" directories that we should exclude from the backup besides just the Bitcoin and Litecoin ones.


backup.sh
```diff
@@ -78,7 +78,7 @@ else
     btcpay_down
 
     echo "Backing up files …"
-    tar --exclude="$backup_dir/*" --exclude="$volumes_dir/generated_bitcoin_datadir/*" --exclude="$volumes_dir/generated_litecoin_datadir/*" --exclude="$volumes_dir/**/logs/*" -cvzf $backup_path $dbdump_path $volumes_dir
+    tar --exclude="$backup_dir/*" --exclude="$volumes_dir/generated_*" --exclude="$volumes_dir/**/logs/*" -cvzf $backup_path $dbdump_path $volumes_dir
```
bolatovumar, line 81: This should exclude all folders which start with "generated_" from the backup. This should cover all blockchain data from all coins that someone might be using. Please let me know if there are reasons why we wouldn't want to do this.

Kukks (approved):
ACK for now, but maybe the correct fix is to apply some labels to volumes to "group them". Then we can label some containers as "backup" and "blockchain" or something

https://stackoverflow.com/questions/69750195/adding-a-label-to-a-named-volume-in-docker-compose

dennisreimann:
Can we make this part of the new backup process? See #641

bolatovumar:
@pavlenex I think we can merge this.

dennisreimann:
I think we should do it in the way that kukks proposed and apply this to the new [backup & restore](https://docs.btcpayserver.org/Docker/backup-restore/) process. We deprecated the `backup.sh` and should not change it anymore.

NicolasDorier:
I agree with @Kukks way as well. However, backups should be opt-out: If a volume doesn't specify a "backup group", then it should be considered to be in all backups.

NicolasDorier:
I will close this PR, I think the `backup.sh` shouldn't be updated. Maybe even removed soon.
In the new system we should definitely have the concept of backup group.

PR #654, opened 2022-05-27
Update Monero Docker links

sethforprivacy:
Simply update the Monero Docker links and Github links to the proper ones.


NicolasDorier:
1. this is auto generated by https://github.com/btcpayserver/btcpayserver-docker/blob/master/contrib/DockerFileBuildHelper/Program.cs
2. I don't think we use the official image, @Kukks may know more about it

sethforprivacy:
Ah, will close this out then, thought manually updating would be useful as it's pointing to unused Dockerfile source.

Feel free to re-open if there is a need for manual updates here.

PR #657, opened 2022-06-07
Fix broken local link that worked only in btcpayserver-doc repo.

ndeet:
ping @dennisreimann


dennisreimann (approved):
Thanks for fixing!

PR #658, opened 2022-06-07
Fix more broken links of old BackupRestore location.

ndeet:
ping @dennisreimann


dennisreimann (approved):
Thanks again for taking care here!

PR #682, opened 2022-07-16
FastSync sign: petzsch

petzsch:


petzsch:
created a keybase profile for this: have it linked up with my existing gpg key, let me know if this works for you.

PR #691, opened 2022-08-08
Fix anonymous datadirs

dennisreimann:
Uses the same folders as in the Dockerfiles to prevent the creation of anonymous volumes as [described here](https://stackoverflow.com/a/62654594/183537).

This should not be merged before someone running Monero in production has reviewed it. I'm not sure what the implications of these changes might be in case the containers/volumes already exist.

### DIFF
--- a/docker-compose-generator/docker-fragments/monero.yml
+++ b/docker-compose-generator/docker-fragments/monero.yml
@@ -10,6 +10,7 @@ services:
       - "18081"
     volumes:
       - "xmr_data:/home/monero/.bitmonero"
+      - "xmr_wallet:/wallet"
   monerod_wallet:
     restart: unless-stopped
     container_name: btcpayserver_monero_wallet
@@ -18,6 +19,7 @@ services:
     expose:
       - "18082"
     volumes:
+      - "xmr_data:/home/monero/.bitmonero"
       - "xmr_wallet:/wallet"
     depends_on:
       - monerod
--- a/docker-compose-generator/docker-fragments/opt-add-lightning-terminal.yml
+++ b/docker-compose-generator/docker-fragments/opt-add-lightning-terminal.yml
@@ -9,7 +9,7 @@ services:
     expose:
       - "8080"
     volumes:
-      - "lnd_lit_datadir:/root"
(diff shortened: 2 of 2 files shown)


Kukks:
cc @sethforprivacy

sethforprivacy:
Will review later today!

sethforprivacy:
Just for clarification -- there is no need to worry about retaining these anon volumes, as the only used directories are the ones already mapped in the `monero.yml` file.

You can see below that both anon volumes are empty and unused, they are simply mapped because we use a unified Docker image for both binaries:

```
sethforprivacy in apps in btcpayserver-docker🔒 on  master ❯ sudo ls -la /var/lib/docker/volumes/6500687d0a1072e79e50a4e7e46dc82f480f74c844fee311046674be67227bf5/_data
total 8
drwxr-xr-x 2 systemd-timesync systemd-timesync 4096 Aug 11 13:48 .
drwx-----x 3 root             root             4096 Sep  1 06:12 ..
sethforprivacy in apps in btcpayserver-docker🔒 on  master ❯ sudo ls -la /var/lib/docker/volumes/0b20801392cc17b9ad814768a534820ad00fc9635f570c84e23b8dc1875468ab/_data
total 8
drwxr-xr-x 2 systemd-timesync systemd-timesync 4096 Aug 11 13:48 .
drwx-----x 3 root             root             4096 Sep  1 06:12 ..
```

sethforprivacy:
I would recommend leaving these unmapped personally, but if you want to make them named volumes regardless I can test on my own instance.

dennisreimann:
This was reported by a user who noticed, that with every backup run (which stops and afterwards restarts the containers) new anonymous data directories get created. As all of them end up being empty, it just spams the volumes – so if possible we should get rid of the problem. 

From what I've seen the added mapping in this PR prevents those additional volumes from being created – but I'm not running this fragment in production, so I cannot say whether or not it might have negative implications.

sethforprivacy:
Interesting! In that case I'll go ahead and test, I don't expect it to cause any issues since they're unused but will report back Monday at the latest.

sethforprivacy:
Tested locally with no issues, LGTM!

PR #706, opened 2022-10-01
Backup: Add optional MariaDB

dennisreimann:
Tested backup and restore on an instance of mine and it works for me, but I wouldn't mind someone else also taking a look.

This is mostly related to the WooCommerce fragment, so maybe @ndeet can help review this. 

In case the MariaDB container is present it gets dumped like the Postgres database container, restore checks if the MariaDB backup is present and restores it too.

I also verified this works on instances without the MariaDB container.

### DIFF
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -28,10 +28,10 @@ fi
 . "$BASH_PROFILE_SCRIPT"
 
 docker_dir=$(docker volume inspect generated_btcpay_datadir --format="{{.Mountpoint}}" | sed -e "s%/volumes/.*%%g")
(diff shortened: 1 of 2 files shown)


ndeet:
Cheers, I actually need to move a testnet VM which runs a WP demo from Hetzner to a new provider. So I will test that with my instance the next couple of days. 👍

ndeet:
I moved a testnet installation successfully from one VM to another VM including wordpress. Great work 👏 

Will do a PR to make backup script not backup testnet blocks and other not needed stuff.

ndeet (approved):

PR #712, opened 2022-10-19
Exclude _data of coins; adding Liquid, Dash, Monero.

ndeet:
Instead of excluding `blocks`, `chainstate` and `indexes` and logs inside the coins `_data` dir we just skip the whole `_data` dir. As we can also remove `_data` when we want to force chain resync I think it should be safe.

This also fixes a bug when you used bitcoin testnet as the exclude pattern would not match anymore on the testnet data dir which is `_data/testnet3/blocks` etc.

I also added Liquid, Monero, Dash which seem to be the only ones popping up on Mattermost and should cover 99% of users. I tested Liquid (elements) and know it works. For Monero and Dash I did not test or double check but assumed the paths from their volumes defined in docker compose fragments. I assumed both also have a `_data` subdirectory.

### DIFF
--- a/btcpay-backup.sh
+++ b/btcpay-backup.sh
@@ -96,14 +96,11 @@ echo "ℹ️ Archiving files in $(pwd)…"
 {
   tar \
     --exclude="volumes/backup_datadir" \
-    --exclude="volumes/generated_bitcoin_datadir/_data/blocks" \
-    --exclude="volumes/generated_bitcoin_datadir/_data/chainstate" \
-    --exclude="volumes/generated_bitcoin_datadir/_data/indexes" \
-    --exclude="volumes/generated_bitcoin_datadir/_data/debug.log" \
-    --exclude="volumes/generated_litecoin_datadir/_data/blocks" \
-    --exclude="volumes/generated_litecoin_datadir/_data/chainstate" \
-    --exclude="volumes/generated_litecoin_datadir/_data/indexes" \
-    --exclude="volumes/generated_litecoin_datadir/_data/debug.log" \
+    --exclude="volumes/generated_bitcoin_datadir/_data" \
+    --exclude="volumes/generated_litecoin_datadir/_data" \
+    --exclude="volumes/generated_elements_datadir/_data" \
+    --exclude="volumes/generated_xmr_data/_data" \
+    --exclude="volumes/generated_dash_datadir/_data" \
     --exclude="volumes/generated_mariadb_datadir" \
     --exclude="volumes/generated_postgres_datadir" \
     --exclude="volumes/generated_electrumx_datadir" \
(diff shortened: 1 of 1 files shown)


dennisreimann:
As long as those coins have a separate volume for the wallets it should be safe to exclude them here. 

I haven't looked at the docker fragments yet, but we should ensure that. Maybe @astupidmoose can help double-check this.

astupidmoose:
Most coins just have whatever bitcoin has in place: Dash, Dogecoin, Litecoin etc. 

XMR is a bit different, but has two separate directories:
generated_xmr_data
generated_xmr_wallet

No idea about liquid as I don't run it.

ndeet:
@astupidmoose thank you for confirming, could you run a command to see if there is a `_data` subdirectory inside?
```
ls -l /var/lib/docker/volumes/generated_xmr_data/
``` 

For Liquid I have already double checked and successfully used the backup + restore to move to a new server.

dennisreimann:
Just checked: [Litecoin](https://github.com/btcpayserver/btcpayserver-docker/blob/master/docker-compose-generator/docker-fragments/litecoin.yml) and [Dash](https://github.com/btcpayserver/btcpayserver-docker/blob/master/docker-compose-generator/docker-fragments/dash.yml) do not seem to have separate volumes for wallets. 

@NicolasDorier Might this be a problem or don't they have (hot) wallets that might live in the data directory?

ndeet:
@dennisreimann the docker compose don't tell the whole truth. I just spun up a VPS to verify and all seem to have a _data subdir, what varies is e.g. for xmr the parent data path which you see in the docker compse files.

```
root@pathtest:~# ls -lah /var/lib/docker/volumes/generated_litecoin_datadir/
total 12K
drwx-----x  3 root root 4.0K Oct 26 13:07 .
drwx-----x 26 root root 4.0K Oct 26 13:07 ..
drwxr-xr-x  2 root root 4.0K Oct 26 13:07 _data
root@pathtest:~# ls -lah /var/lib/docker/volumes/generated_dash_datadir/
total 12K
drwx-----x  3 root root 4.0K Oct 26 13:07 .
drwx-----x 26 root root 4.0K Oct 26 13:07 ..
drwxr-xr-x  2 root root 4.0K Oct 26 13:07 _data
root@pathtest:~# ls -lah /var/lib/docker/volumes/generated_xmr_data/
total 12K
drwx-----x  3 root root 4.0K Oct 26 13:07 .
drwx-----x 26 root root 4.0K Oct 26 13:07 ..
drwxr-xr-x  2 root root 4.0K Oct 26 13:07 _data
root@pathtest:~# ls -lah /var/lib/docker/volumes/generated_dogecoin_datadir/
total 12K
drwx-----x  3 root             root             4.0K Oct 26 13:11 .
drwx-----x 29 root             root             4.0K Oct 26 13:11 ..
drwxr-xr-x  6 systemd-coredump systemd-coredump 4.0K Oct 26 13:11 _data
```

dennisreimann:
@ndeet the `_data` subdir is normal, all the directories have it. What I meant is that we need to check whether or not wallet files for those altcoins are in those directories.

ndeet:
ok, I see what you meant now. Would be pita to sync all those shitcoins so hope @NicolasDorier remembers, we can ask him in our call now. If he does not remember I can spin up a shtcoin vm with litecoin and dash again and let them sync and create a wallet.

NicolasDorier:
we don't have to care too much about shitcoin, doing it as a best effort manner.

PR #731, opened 2022-11-27
Move Traefik labels to services files

iBobik:
It already happened in the past that people updating services yamls forgot about Traefik and broke it, updating only Nginx config. Good practice to mitigate it is to have the config nearby, like currently with Nginx env variables.

There was [some issues with some users](https://github.com/btcpayserver/btcpayserver-docker/pull/585#issuecomment-1159607741), @NicolasDorier could you please connect me with them, so we can test it better? It should work according to docs.

If not, I would try to upgrade compose schema version from 3 to at least 3.7.


iBobik:
Still works well in my deployment. What is against merging it?

Kukks (approved):
Sorry for taking o long, i didnt realize this PR was here.

PR #735, opened 2022-12-06
Bump clightning

NicolasDorier:


NicolasDorier:
Need to wait next update of btcpay for merging that. The public node info doesn't show the address because of a breaking change.

NicolasDorier:
So updating c-lightning on my server bricked it. The sync was stuck. I will try to see if I can repro of it it was me just updating a node that didn't run for month.

PR #736, opened 2022-12-07
Activate mempoolfullrbf on BTCPay deployments

NicolasDorier:
# Explanation

Mempool Full RBF is a mempool policy.
If this policy is turned on, when a new transaction is considered to be added to mempool and a conflict (double spending) is detected:
* If the new transaction has higher fee, then the previous transaction get replaced (in any case)
* Else the the new transaction is ignored

Without this policy,
* If the new transaction has higher fee AND the previous transaction signal RBF, then the previous transaction get replaced.
* Else the the new transaction is ignored

Bitcoin Core 24.0 allows nodes to activate this policy, however, it is turned off by default.

# Rational

We should activate `mempoolfullrbf=1`, because as a merchant you want to know if a payment is getting double spent as soon as possible.

Without `mempoolfullrbf=1` if a customer is double spending a payment with a transaction not signaling RBF, the merchant would only know after confirmation.

However, some users doesn't want to activate this policy, as some people consider that while it aligns with the incentive of the merchant to activate it, it is considered against the interests of the network, as it make accepting a payment with 0 confirmations harder once the policy is widely deployed.

For this reason, people who wants to disable this policy may opt out with

```bash
BTCPAYGEN_EXCLUDE_FRAGMENTS="$BTCPAYGEN_EXCLUDE_FRAGMENTS;opt-mempoolfullrbf"
. btcpay-setup.sh -i
```

### DIFF
--- a/docker-compose-generator/docker-fragments/bitcoin.yml
+++ b/docker-compose-generator/docker-fragments/bitcoin.yml
@@ -42,3 +42,5 @@ volumes:
 
 exclusive:
   - bitcoin-node
+recommended:
+  - "opt-mempoolfullrbf"
--- a/docker-compose-generator/docker-fragments/opt-mempoolfullrbf.yml
+++ b/docker-compose-generator/docker-fragments/opt-mempoolfullrbf.yml
@@ -1,0 +1,8 @@
+version: "3"
+#  must not use opt-save-storage
+
+services:
+  bitcoind:
+    environment:
+      BITCOIN_EXTRA_ARGS: |
+        mempoolfullrbf=1


Kukks (changes requested):
Considering this is a potentially controversial decision, I would suggest we put it in a fragment but activated by default through a recommended yml link. This way, users can opt out easily.

petertodd:
FYI I measured full-rbf adoption among Bitcoin core v24 nodes listening on IPv4, and found that at least 17% have enabled full-rbf and are succesfully propagating replacements: https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-December/021296.html

An argument against enabling full-rbf in this circumstance would be keeping it disabled would materially reduce the chances of full-rbf replacements from propagating. But between that wide adoption, and the existence of full-rbf peering nodes (Bitcoin Knots and [my own fork](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-December/021270.html)), I don't think this argument holds any more. It's also a fact that the mining pool Luxor - who has enabled full-rbf - is consistently getting full-rbf replacements without propagation issues.

So ACK: you might as well know.

NicolasDorier:
@Kukks I changed this PR so mempoolfullrbf is a separate recommended fragment.
Meaning that it will be automatically on, but people can still disable `mempoolfullrbf` with

```bash
BTCPAYGEN_EXCLUDE_FRAGMENTS="$BTCPAYGEN_EXCLUDE_FRAGMENTS;opt-mempoolfullrbf"
. btcpay-setup.sh -i
```

PR #739, opened 2022-12-16
Adding steps for a server migration.

ndeet:


docs/backup-restore.md
```diff
@@ -175,3 +175,56 @@ Always make sure your backup strategy is tested and fits your needs.
 No one solution fits all, and we tried to cover the basic cases.
 For the latest updates, always feel free to ask on the BTCPay Server community channels.
 :::
+
+## How to migrate to a new server?
+
+The necessary steps always depend on your setup and infrastructure but as a rough guideline the following steps and considerations will get you through the process.
+
+:::warning
+If you have Lightning Network (LN) channels open on your internal node, you need to make sure to **not** restore any old channel state, otherwise your are at the risk of loosing funds. Therefore it is important to shut down BTCPay on the old server immediately after the backup has finished as outlined below.
```
dennisreimann, line 184:
```suggestion
If you have Lightning Network (LN) channels open on your internal node, you need to make sure to **not** restore any old channel state, otherwise your are at the risk of losing funds.
Therefore it is important to shut down BTCPay Server on the old server immediately after the backup has finished as outlined below.
```

docs/backup-restore.md
```diff
@@ -175,3 +175,56 @@ Always make sure your backup strategy is tested and fits your needs.
...
+:::
+
+
+### Preparation   
+
+#### Temporarily reduce the TTL of your domain:   
+Depending on the TTL (time-to-live) of your DNS entries (A-Record) for you BTCPay Server domains/subdomains it makes sense to reduce the TTL a day or a few hours before you want to start the migration. In case anything goes wrong you can quickly switch between new and old server IP addresses. You should set the TTL for the domain/subdomain to some shorter timewindow eg. 600 seconds (10 minutes) or if possible 300 seconds (5 mins). You can do this in at your domain registrar control center.
```
dennisreimann, line 191:
```suggestion
Depending on the TTL (time-to-live) of your DNS entries (A-Record) for you BTCPay Server domains/subdomains it makes sense to reduce the TTL a day or a few hours before you want to start the migration.
In case anything goes wrong you can quickly switch between new and old server IP addresses.
You should set the TTL for the domain/subdomain to some shorter time frame eg. 600 seconds (10 minutes) or if possible 300 seconds (5 mins).
You can do this in at your domain registrar control center.
```

docs/backup-restore.md
```diff
@@ -175,3 +175,56 @@ Always make sure your backup strategy is tested and fits your needs.
...
+
+#### Environment variables:
+write down the environment variables you need for the setup on the new server as written in the [Docker installation guide](/Docker/#full-installation-for-technical-users).
```
dennisreimann, line 194:
```suggestion
Write down the environment variables you need for the setup on the new server as written in the [Docker installation guide](/Docker/#full-installation-for-technical-users).
```

docs/backup-restore.md
```diff
@@ -175,3 +175,56 @@ Always make sure your backup strategy is tested and fits your needs.
...
+echo $NBITCOIN_NETWORK
+echo $BTCPAYGEN_CRYPTO1
+echo $BTCPAYGEN_ADDITIONAL_FRAGMENTS
+echo $BTCPAYGEN_REVERSEPROXY
+echo $BTCPAYGEN_LIGHTNING
+echo $BTCPAY_ENABLE_SSH
+```
+
+:::tip
+If you have custom modifications or additional altcoins then do not forget write down those environment variables too.
```
dennisreimann, line 207:
```suggestion
If you have custom modifications or additional altcoins then do not forget to write down those environment variables too.
```

dennisreimann (commented):
I'd say the installation on the new server could happen before the (DNS) switch, so that only the restore process has to be run as the final step.

ndeet:
Thanks a lot for your fixes 💚 

> I'd say the installation on the new server could happen before the (DNS) switch, so that only the restore process has to be run as the final step.

I was not sure if the letsencrypt certs nginx vhosts etc. will work with the backup & restore. But thinking about it if everything is copied over they should and there is no need to have a new cert to be issued on the new server 🤔 ... what will happen in that case though is that the letsencrypt cert creation attempt will fail and show some error during installation? Not sure what is better.

dennisreimann:
Hmm, good point, I haven't tried either — I think the cert creation just fails without interupting the setup and it'll try again later.

ndeet:
Yes, it for sure does not interrupt the setup but likely shows some error messages which might confuse users or if everything is alright.

dennisreimann (approved):

PR #740, opened 2022-12-21
Bump Torq to v0.15.4

maxwedwards:
Apologies that we only released a week ago but we got a few nice bug fixes in and a new network toggle for changing between mainnet and the various test nets which keeps the data segregated for those who use Torq with a testnet node as well as a main net node.

Tested on a real BTC Pay Server node.


dennisreimann (approved):
No worries, feel free to update whenever you want.

NicolasDorier:
@dennisreimann if the maintainer of a project open a PR like this and you see it, just merge. We don't have to test ourselves.

PR #746, opened 2023-01-05
Update petzsch-utxo-sets.asc

petzsch:
verified and signed


dennisreimann (approved):
Verified hash and signature.

PR #747, opened 2023-01-09
Allow NBX to provide rpc proxy

Kukks:
This allows btcpay to use the node rpc directly instead of needing direct rpc access. Should not be any security implications since nbx is still an internal service.


NicolasDorier:
I prefer leaving off, but since the methods you need aren't dangerous, I will expose those even if EXPOSERPC is 0.
The advantage is that integrators don't have to think to have a new flag.

https://github.com/dgarage/NBXplorer/pull/396

Monicaann0o:
I appreciate you"!
Monica Ann Holguin
323)309-2366
***@***.***

PR #775, opened 2023-04-15
bump mempool to v2.5.0

petzsch:
2 Open Questions/Points before merging:

1. According to [release notes](https://github.com/mempool/mempool/releases/tag/v2.5.0) this version comes with a lightning network explorer. Not sure if that requires config changes to the docker-compose fragment so that LND/C-Lightning can be accessed. (we might have opted out of this possibility before for whatever reason)

2. Should we also bump [MariaDB Image](https://github.com/mempool/mempool/releases/tag/v2.5.0) used in this setup? We deliver 10.5.8 - maybe we want to pin it to 10?


dennisreimann:
1. I didn't find specific docs for the LN explorer. Judging from the PR it [looks simple enough](https://github.com/mempool/mempool/pull/1976/files#diff-11a7c31e37dbd48f3baad47ffb9227445e72f567b1699bf0ee59dab34dce707c) to give it a try, the necessary values are exposed already I think. Associated mempool config variables:
```
  "LIGHTNING": {
    "ENABLED": false,
    "BACKEND": "lnd"
  },
  "LND": {
    "TLS_CERT_PATH": "tls.cert",
    "MACAROON_PATH": "admin.macaroon",
    "SOCKET": "localhost:10009"
  }
```

2. I'd be fine with bumping/pining a more generic version — haven't found anything in terms of specific version requirements from the Mempool side either.

petzsch:
would be cool if the lightning part would only get activated if btcpayserver has a comptaible Lightning backend (CLN or LND) running.

That way non-lightning users could still use the mempool interface.

I'll put some more thought into this, but I guess might need some help with the fragments syntax. The not so elegant way would probably be to create 3 seperate mempool fragments (no-lightning, CLN, LND) and have the user configure the one he wants.

petzsch:
@NicolasDorier Could you check commit https://github.com/btcpayserver/btcpayserver-docker/pull/775/commits/19c5277e714ee97c2456313ede432c0c68cf070a if opening up permissions on the lightning-rpc socket could have security implications? (feels like a change where some more eyes are needed)

I've tried accessing it from outside docker as a regular "ubuntu" user und wasn't able to see anything in /var/lib/docker/ ... so I think we should be save even on systems running other stuff then our deployment.

The alternative would be to run mempool-backend container as user root inside and not as user "node".

NicolasDorier:
I think this change is fine. We can merge this if you tested it and made sure to not brick existing install. (the db migration is only a minor release? if it's major sometimes it's more difficult than just bumping image version)

petzsch:
### tested:

- [X] CLN+mempool
- [X] LND+mempool
- [X] CLN
- [X] LND
- [X] mempool w/o lightning*

\* When unsetting BTCPAYGEN_LIGHTNING and configuring the opt-add-mempool fragment where just LND was priviously running, I crashed my nginx with this in the logs after running . ./btcpay-setup.sh -i: 

```
2023/04/26 07:45:57 [emerg] 1#1: host not found in upstream "bitcoin_rtl" in /etc/nginx/conf.d/default.conf:92
nginx: [emerg] host not found in upstream "bitcoin_rtl" in /etc/nginx/conf.d/default.conf:92
```

A btcpay-restart.sh seems to have fixed the issue.

### Regarding MariaDB:

Since both are 10.x versions of MariaDB we should be fine. I upgraded my old mempool install to this one and didn't run into any database problems. But yeah once 11 becomes stable, we need to look out for breaking changes.

PR #792, opened 2023-05-21
exclude blocks and chainstate from doge backup

petzsch:
A user on TG had a problem with creating a backup of his Altcoin build instance from Lunanode because the backup exceeded the available storage on / partition.

I hope that the missing exclude for dogecoin was the reason, all other alts he uses seem to be in our exclude list.

Would appreciate if this could be merged in a timely manner.
Thanks.

cc @dennisreimann


dennisreimann (approved):
Thanks for the addition!

Issue #805, opened 2023-06-26
monerod crashes after update to Monero v0.18.2.2


d2ro:
Todays commits caused monerod to crash on our server:

```
# docker logs --tail 10 "btcpayserver_monerod" 
2023-06-26 09:49:59.427	I Stopping cryptonote protocol...
2023-06-26 09:49:59.427	I Cryptonote protocol stopped successfully
2023-06-26 09:49:59.428	E Exception in main! boost::filesystem::create_directories: Permission denied: "/nonexistent"
2023-06-26 09:51:01.007	I Monero 'Fluorine Fermi' (v0.18.2.2-release)
2023-06-26 09:51:01.007	I Initializing cryptonote protocol...
2023-06-26 09:51:01.007	I Cryptonote protocol initialized OK
2023-06-26 09:51:01.008	I Initializing core...
2023-06-26 09:51:01.008	I Stopping cryptonote protocol...
2023-06-26 09:51:01.008	I Cryptonote protocol stopped successfully
2023-06-26 09:51:01.008	E Exception in main! boost::filesystem::create_directories: Permission denied: "/nonexistent"
```

We temporarily fixed it by going back to commit `cd0a734`:

```
# cd btcpayserver-docker/
# git checkout cd0a73435c2bc8a8bd7c815611ef34c634ca3628
# btcpay-update.sh --skip-git-pull
```

NicolasDorier:
Ping @sethforprivacy

sethforprivacy:
Thanks, looking now.

I wonder if the mounts are messed up for this specific install? @NicolasDorier do you know what would cause a container in btcpay to try and mount `/nonexistent`?

sethforprivacy:
Reproduced the issue on 1.10.2 w/ the new 0.18.2.2 Monero image.

sethforprivacy:
I think I have a fix identified, testing now, then will PR.

sethforprivacy:
https://github.com/btcpayserver/dockerfile-deps/pull/67 resolves this, it appears to be a permissions issue as the containers default UID for the new Monero user defaults to 100 now instead of 101 after switching the base image to `stable-slim`.

NicolasDorier:
@sethforprivacy  `adduser: The --group, --ingroup, and --gid options are mutually exclusive.`

sethforprivacy:
> @sethforprivacy  `adduser: The --group, --ingroup, and --gid options are mutually exclusive.`

[...]

PR #815, opened 2023-07-09
Fix Monero image tag

sethforprivacy:
It seems that https://github.com/btcpayserver/dockerfile-deps/commit/1d7b2a41909f59da5bcc14237d59bcda1ff64f56 fixed Monero, but the tag was improperly referenced here as `-4` when the folder is simply `0.18.2.2` in `dockerfile-deps`:

https://github.com/btcpayserver/dockerfile-deps/tree/master/Monero/0.18.2.2

I tested the Monero container on it's own from https://github.com/btcpayserver/dockerfile-deps/commit/1d7b2a41909f59da5bcc14237d59bcda1ff64f56 and it seems to use the proper user and group and work to the best of my abilities to test.

### DIFF
--- a/docker-compose-generator/docker-fragments/monero.yml
+++ b/docker-compose-generator/docker-fragments/monero.yml
@@ -4,7 +4,7 @@ services:
   monerod:
     restart: unless-stopped
     container_name: btcpayserver_monerod
-    image: btcpayserver/monero:0.18.2.2-4
+    image: btcpayserver/monero:0.18.2.2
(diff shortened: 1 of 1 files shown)


sethforprivacy:
Ping @Kukks @NicolasDorier, I think you resolved https://github.com/btcpayserver/btcpayserver-docker/issues/805 and just had missed that the folder structure was never updated to use the `-4` postfix.

I corrected that here.

NicolasDorier:
@sethforprivacy I checked the `-4` is correct. There is no way to replace a monero image that was already published while asking users to update, so we manually apply revision information.

PR #825, opened 2023-08-20
Update opt-add-thunderhub.yml

apotdevin:


dennisreimann:
If you pin the sha this will lead to problems with the different architectures, at least I've seen problems like this before. ThunderHub supports all archs, right?

apotdevin:
@dennisreimann it supports amd64, armv7 and arm64

It is pinned to the multiarch tag rather than a specific arch so it should work fine for all :)

dennisreimann (approved):

PR #843, opened 2023-11-01
Add Fulcrum

petzsch:
I would like to add the [Fulcrum](https://github.com/cculianu/Fulcrum) explorer as an option/alternative to electrumX.

However I don't know how to make this run with the existing mempool.space config. Hence I created a customized copy of the same to work with Fulcrum.

The current config is running on one of my lab nodes with pretty impressive performance.

I'm putting this here as Draft in the hope that someone has an idea how to fix my issue.

I think we also still need to add the docker image creation steps to the build all images script. And update the Readme accordingly.

### DIFF
--- a/docker-compose-generator/docker-fragments/opt-add-fulcrum.yml
+++ b/docker-compose-generator/docker-fragments/opt-add-fulcrum.yml
@@ -1,0 +1,23 @@
+version: "3"
+
+services:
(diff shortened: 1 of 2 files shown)


petzsch:
outdated and breaks compatibility with mempool fragment -> giving up on this one.

PR #865, opened 2023-12-26
Move from python-teos to rust-teos

benjaminchodroff:
I am not the owner of TEOS, but hope that we can see BTCPayServer begin to move away from the archived python-teos to the new rust-teos. This pull request does not install the plugin to CLN, but does at least get the TEOS server working as expected. 

Installing the CLN client plugin requires additional manual setup. If others are interested, I'll try to document those steps.


benjaminchodroff:
Happy to close this pull request if there is a better image to use from talaia-labs.

g00g1:
New version does not require txindex: https://github.com/talaia-labs/rust-teos/commit/221358d6d453670c4fc3a1bf5d651db2c21930cb

Kukks (approved):

Issue #869, opened 2024-01-10
last upgrade crashed on debian bookworm


anarcat:
Our latest upgrade crashed because btcpayserver seems to think our docker install ([20.10.24 in Debian bookworm](https://tracker.debian.org/pkg/docker.io)) is too old. And the way it seems to handle this is ... rather sub par:

```
Subject: Cron <root@btcpayserver-02> /root/BTCPayServer/btcpayserver-docker/btcpay-update.sh

From https://github.com/btcpayserver/btcpayserver-docker
   6b5841e..4e8aa5e  master            -> origin/master
   e265d0d..aa55e7c  dcg-latest        -> origin/dcg-latest
 * [new branch]      joinmarket-0.9.10 -> origin/joinmarket-0.9.10
Updating 6b5841e..4e8aa5e
Fast-forward
 Production/nginx.tmpl                              |  19 +-
 README.md                                          |  23 +-
 btcpay-setup.sh                                    |  11 -
 btcpay-update.sh                                   |   5 +-
 contrib/DockerFileBuildHelper/Program.cs           |  27 ++-
 contrib/FastSync/README.md                         |   2 +-
 contrib/FastSync/load-utxo-set.sh                  |   2 +-
 contrib/FastSync/sigs/NicolasDorier.utxo-sets.asc  |  27 +--
 contrib/FastSync/utxo-sets                         |   3 +-
 contrib/build-all-images.sh                        | 238 ++++++++++-----------
 .../docker-fragments/bitcoin-lnd.yml               |   3 +-
 .../docker-fragments/bitcoin.yml                   |   2 +-
 .../docker-fragments/btcpayserver.yml              |   3 +-
 docker-compose-generator/docker-fragments/dash.yml |   2 +-
 .../docker-fragments/nbxplorer.yml                 |   4 +-
 .../docker-fragments/nginx-https.yml               |   6 +-
 .../docker-fragments/nginx.yml                     |   4 +-
 .../docker-fragments/opt-add-cloudflared.yml       |   2 +-
 .../docker-fragments/opt-add-fireflyiii.yml        |   1 +
 .../docker-fragments/opt-add-tallycoin-connect.yml |   1 +
 .../docker-fragments/opt-add-thunderhub.yml        |   3 +-
 .../docker-fragments/opt-add-tor-relay.yml         |   4 +-
 .../docker-fragments/opt-add-tor.yml               |   4 +-
[...]

PR #870, opened 2024-01-10
add a way to skip docker updates altogether

anarcat:
One way of fixing #869.

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -160,6 +160,9 @@ docker_compose_update() {
 }
 
 docker_update() {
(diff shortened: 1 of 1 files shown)


Kukks (changes requested):
This would require a gitignore addition as it would stop future updates to the repo entirely (btcpay-update does a git pull)

Although, I would prefer an env var being persisted with the rest of our settings for this.

anarcat:
i think there's a `HAS_DOCKER` variable now that might do what i need.

PR #883, opened 2024-02-29
make BTCPAY_ENV_FILE changeable by user

TheKhanj:
It would be nice to change env file location.


Kukks (changes requested):
Wouldn't we end up overwriting this file at some point with the current changes?

PR #941, opened 2024-10-18
Bump RTL to 0.15.2

rockstardev:
Bumping RTL to 0.15.2 to get on the latest release


rockstardev:
seems we can't bump RTL version, there is a configuration error for 0.15.2

image

rockstardev:
OK, so I had an instance on RTL 0.14.1 that was working without problems, bumped it to 0.15.2 and now I am getting the same error:

image

rockstardev:
Resolved eventually through: #956

PR #956, opened 2024-11-19
Bumping RTL to v0.15.4

ShahanaFarooqui:


rockstardev:
I've tested it on my node and can finally access the dashboard!

However, there's a small quirk with the login due to changed behavior. If I access the link once, any subsequent clicks on the link will fail with the error: `SSO Authentication Failed! Access key too short or does not match.` Here's a video illustrating the issue:

https://github.com/user-attachments/assets/9a0d16c5-e96f-471d-a1e4-9b38e0fd580a

This issue is likely related to the authentication key. My suggestion is to leave it as is for now. Alternatively, we could update this on the BTCPay Server side to refresh the link once it's clicked.

ShahanaFarooqui:
> However, there's a small quirk with the login due to changed behavior. If I access the link once, any subsequent clicks on the link will fail with the error: `SSO Authentication Failed! Access key too short or does not match.` Here's a video illustrating the issue:
> 
> This issue is likely related to the authentication key. My suggestion is to leave it as is for now. Alternatively, we could update this on the BTCPay Server side to refresh the link once it's clicked.

This issue has persisted since we began refreshing the cookie in RTL after each use. It has been a longstanding problem for at least past four years. While I could remove the cookie refresh feature from RTL, ideally the link should be updated on the BTCPay Server page after it is clicked.

PR #957, opened 2024-11-21
Updating FastSync UTXO set and location of it

rockstardev:
It's been a year since @NicolasDorier last updated UTXO set, providing the one more recent and ensuring it's on hosting that @ndeet brought up.


contrib/FastSync/load-utxo-set.sh
```diff
@@ -39,7 +39,7 @@ fi
 TAR_FILE="$1"
 
 if ! [[ "$UTXO_DOWNLOAD_LINK" ]]; then
-    [[ $NBITCOIN_NETWORK == "mainnet" ]] && UTXO_DOWNLOAD_LINK="https://eu2.contabostorage.com/1f50a74c9dc14888a8664415dad3d020:utxosets/utxo-snapshot-bitcoin-mainnet-820852.tar"
+    [[ $NBITCOIN_NETWORK == "mainnet" ]] && UTXO_DOWNLOAD_LINK="https://snapshots.btcpay.tech/utxo-snapshot-bitcoin-mainnet-867690.tar"
```
petzsch, line 42:
```suggestion
    [[ $NBITCOIN_NETWORK == "mainnet" ]] && UTXO_DOWNLOAD_LINK="https://fastsync.btcpay.host/utxo-snapshot-bitcoin-mainnet-867690.tar"
```

This uses my mirrorselect/GeoIP script to 301 Redirect to the nearest mirror (currently only Helsinki (Hetnzer = ndeet) and Netcup.eu (Vienna) are in my list.

Would be nice if you could consider the change. This allows adding of new mirrors by me simply adding them in my mirror list.

I can provide you with the PHP source of the redirect script if you want to host it yourself - fine with me.

petzsch (changes requested):
Please consider alternative mirror management approach suggest in my comment.

Thx.

contrib/FastSync/load-utxo-set.sh
ndeet: Having a mirror script with geo loctation might be not a bad idea. Can you share the script so we can check and decide where to host and if it makes sense. Or we leave it as is for this PR but open another one to introduce and discuss how to approach the mirroring?

petzsch: Here is the script if you wanna host it yourselves: https://github.com/petzsch/btcpay-geoip-fastsync

rockstardev: Guys, let's not complicate this PR with geo script. If you want to work on this separately - open an issue after this is reviewed and merged.

rockstardev:
@petzsch thanks for offering, but let's keep it simple and within the hosting that the core team has access to. After this is updated, we can consider opening a new issue if there is long-term will to support this through some alternative.

PR #974, opened 2025-02-14
Add Litecoin MWEB docker fragment

hectorchu:
This works in combination with the new Litecoin MWEB payment handler plugin.


NicolasDorier:
I'd advice you to put some docs on README

PR #977, opened 2025-02-19
Add shopify app deployer plugin

NicolasDorier:


docker-compose-generator/docker-fragments/opt-add-shopify.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+services:
+  shopify-app-deployer:
+    image: btcpayserver/shopify-app-deployer:1.0
```
ndeet, line 4: tag should be 1.1?

PR #987, opened 2025-04-25
Add Phoenixd support

pm47:
This adds `phoenixd` as an official Lightning implementation:
```shell
export BTCPAYGEN_LIGHTNING="phoenixd"
```

We already have a working plugin, will publish shortly.

### DIFF
--- a/docker-compose-generator/docker-fragments/phoenixd.yml
+++ b/docker-compose-generator/docker-fragments/phoenixd.yml
@@ -1,0 +1,22 @@
+version: "3"
+
+services:
+  phoenixd:
+    image: acinq/phoenixd:0.5.1
+    container_name: phoenixd
+    restart: unless-stopped
+    networks:
+      - default
+    command: ["--chain=${NBITCOIN_NETWORK:-regtest}"]
+    expose:
+      - "9740"
+    depends_on:
+      - btcpayserver
+    volumes:
+      - "phoenixd_datadir:/phoenix"
+
+volumes:
+  phoenixd_datadir:
+
+exclusive:
+  - lightning
--- a/docker-compose-generator/crypto-definitions.json
+++ b/docker-compose-generator/crypto-definitions.json
@@ -4,83 +4,95 @@
     "CryptoFragment": "litecoin",
     "CLightningFragment": null,
     "LNDFragment": null,
-    "EclairFragment": null
+    "EclairFragment": null,
+    "PhoenixdFragment": null
   },
   {
     "Crypto": "btc",
     "CryptoFragment": "bitcoin",
     "CLightningFragment": "bitcoin-clightning",
     "LNDFragment": "bitcoin-lnd",
-    "EclairFragment": "bitcoin-eclair"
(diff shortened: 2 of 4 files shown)


docker-compose-generator/docker-fragments/phoenixd.yml
```diff
@@ -0,0 +1,22 @@
...
+    networks:
+      - default
+    command: ["--chain=${NBITCOIN_NETWORK:-regtest}"]
+    expose:
+      - "9740"
+    depends_on:
+      - btcpayserver
+    volumes:
+      - "phoenixd_datadir:/phoenix"
+
```
pm47, line 17: I think we may need this section to define the default connection string, with a hardcoded api password.

```suggestion
  btcpayserver:
    environment:
      BTCPAY_BTCLIGHTNING: "type=phoenixd;server=http://phoenixd:8080;password=xxx"
    volumes:
      - "phoenixd_datadir:/etc/phoenix"
    links:
      - phoenixd
```

rockstardev:
wow, let's go!

NicolasDorier:
@pm47 that's cool. One question: Is phoenixd just a lightning implementation? If so we should allow user to install it instead of LND/CLN and also develop our lightning client provider so people can just use phoenixd as backend for BTCPay.

pm47:
> @pm47 that's cool. One question: Is phoenixd just a lightning implementation? If so we should allow user to install it instead of LND/CLN and also develop our lightning client provider so people can just use phoenixd as backend for BTCPay.

We started with a plugin because it was the easiest way to prototype, but yes it should really be a lightning implementation.

> @pm47 please ping me on mattermost or twitter if you need help. Depending on what your plugin do, we can integrate it in core by just adding your client implementation to https://github.com/btcpayserver/btcpayserver.lightning

Will do thanks!

NicolasDorier:
@pm47 just do your stuff as it seems most convenient to you. I can help you pick it up after. (in plugin or in https://github.com/btcpayserver/btcpayserver.lightning directly)

As you can see there is already an Eclair implementation. Should we remove Eclair support and just have Phoenix support? Or they are both separate?

pm47:
> I can help you pick it up after. (in plugin or in https://github.com/btcpayserver/btcpayserver.lightning directly)

Thanks! went ahead in https://github.com/btcpayserver/BTCPayServer.Lightning/pull/169

> Should we remove Eclair support and just have Phoenix support? Or they are both separate?

They are separate, might as well keep both? Except if you'd rather limit the total number of implementations

docker-compose-generator/docker-fragments/phoenixd.yml
pm47: Done.

NicolasDorier:
@pm47 phoenixd is implemented in btcpay in the next minor release. We can add phoenixd in our deployment docker-compose like now, but can you ping me on twitter or chat.btcpayserver.org? I am trying to understand better how phoenixd work to know if anything else is needed in this PR.

PR #989, opened 2025-05-02
add knots as a bitcoind backend

Retropex:
Hey,

I would need some help with this, I would like that if `export BTCPAYGEN_CRYPTO1="btc-knots"` is exported then knots is utilized as a backend.

The thing I don't understand where in `docker-compose-generator` the `btc`/`btc-knots` is interpreted.

### DIFF
--- a/docker-compose-generator/docker-fragments/bitcoinknots.yml
+++ b/docker-compose-generator/docker-fragments/bitcoinknots.yml
@@ -1,0 +1,46 @@
+version: "3"
+
+services:
+  bitcoind:
+    restart: unless-stopped
+    container_name: btcpayserver_bitcoind
+    image: ghcr.io/retropex/bitcoin:btcpay-28.1
+    environment:
+      BITCOIN_NETWORK: ${NBITCOIN_NETWORK:-regtest}
+      CREATE_WALLET: "false"
+      BITCOIN_WALLETDIR: "/walletdata"
+      # rpcport and rpcbind seems duplicates, but they are not
+      # rpcport is using by some tooling to automatically get
+      # the rpcport from the configuration file. Do not remove!
+      BITCOIN_EXTRA_ARGS: |
+        rpcport=43782
+        rpcbind=0.0.0.0:43782
+        rpcallowip=0.0.0.0/0
+        port=39388
+        whitelist=0.0.0.0/0
+        maxmempool=500
+    expose:
+      - "43782"
+      - "39388"
+    volumes:
+      - "bitcoin_datadir:/data"
+      - "bitcoin_wallet_datadir:/walletdata"
+  nbxplorer:
+    environment:
+      NBXPLORER_CHAINS: "btc"
+      NBXPLORER_BTCRPCURL: http://bitcoind:43782/
+      NBXPLORER_BTCNODEENDPOINT: bitcoind:39388
+    volumes:
+      - "bitcoin_datadir:/root/.bitcoin"
+  btcpayserver:
+    environment:
+      BTCPAY_CHAINS: "btc"
+      BTCPAY_BTCEXPLORERURL: http://nbxplorer:32838/
+volumes:
+  bitcoin_datadir:
+  bitcoin_wallet_datadir:
+
+exclusive:
+  - bitcoin-node
+recommended:
+  - "opt-mempoolfullrbf"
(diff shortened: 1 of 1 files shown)


ndeet:
I doubt replacing btc with btc-knots is the right approach, feels more like adding an altcoin. I looked into adding a custom fragment and replacing the just image but that does not work as our docker generator can only append and not replace existing config in yaml files.

To my surprise I found @kukks already built knots images until 2 years ago on our docker hub https://hub.docker.com/r/btcpayserver/bitcoinknots/tags and https://github.com/btcpayserver/dockerfile-deps/tree/master/BitcoinKnots

So maybe we already have something in place @Kukks? I know for the BTCPay image we can use BTCPAY_IMAGE env variable, do we have something similar to replace the bitcoind image? Can you give some hints?

Kukks:
> I doubt replacing btc with btc-knots is the right approach, feels more like adding an altcoin. I looked into adding a custom fragment and replacing the just image but that does not work as our docker generator can only append and not replace existing config in yaml files.
> 
> To my surprise I found @kukks already built knots images until 2 years ago on our docker hub https://hub.docker.com/r/btcpayserver/bitcoinknots/tags and https://github.com/btcpayserver/dockerfile-deps/tree/master/BitcoinKnots
> 
> So maybe we already have something in place @Kukks? I know for the BTCPay image we can use BTCPAY_IMAGE env variable, do we have something similar to replace the bitcoind image? Can you give some hints?

Yes, I provided alternative clients for during the taproot activation dilemma. Once taproot activated, we removed it (also because it wasn't as generic and setup was very specific, with specific images)

I agree, we can either:
* introduce a BITCOIN_IMAGE env var
* separate the image definition of the bitcoin fragment as a separate fragment (bitcoin-node.yml, with just the image definition) and have bitcoin.yml recommend it. This way a user may just exclude bitcoin-node, and include bitcoinknots-node. This lets the whole build system not need to be changed at all and is generally just modular (at the expense of complexity)

ndeet:
@Kukks thanks a lot for the context and suggestions

@NicolasDorier what way do you think is the best to make this happen?

NicolasDorier:
just do the same stuff we did before

petzsch:
Assuming we are talking about how implementation was done for https://github.com/btcpayserver/btcpayserver-docker/pull/465/files

We would just create a copy of bitcoin.yml called bitcoinknots.yml, modify the image and tell people to add:
```bash
export BTCPAYGEN_EXCLUDE_FRAGMENTS="bitcoin"
export BTCPAYGEN_ADDITIONAL_FRAGMENTS="opt-save-storage-s;bitcoinknots"
```

Tested these changes on my node and it seems to run so far.
Maybe there is a way without having to exclude the bitcoin fragment?

docker-compose-generator/docker-fragments/bitcoinknots.yml
```diff
@@ -0,0 +1,46 @@
+version: "3"
+
+services:
+  bitcoind:
+    restart: unless-stopped
+    container_name: btcpayserver_bitcoind
+    image: ghcr.io/retropex/bitcoin:btcpay-28.1
```
petzsch, line 7:
```suggestion
    image: btcpayserver/bitcoinknots:28.1
```

mp21-org (changes requested):

mp21-org (approved):
looks good to me.

petzsch:
It should probably also be added to this repositories [README.md](https://github.com/btcpayserver/btcpayserver-docker/?tab=readme-ov-file#generated-docker-compose).

mp21-org (changes requested):
.\contrib\DockerFileBuildHelper\update-repo.ps1
or
.\contrib\DockerFileBuildHelper\update-repo.sh

should be run to update the [build all images script](https://github.com/btcpayserver/btcpayserver-docker/blob/master/contrib/build-all-images.sh) and parts of the README.
The description what the fragment does would still need to be added manually to the [README.md](https://github.com/btcpayserver/btcpayserver-docker/?tab=readme-ov-file#generated-docker-compose)

NicolasDorier:
can cleanup the `update-repo.ps1` doc later.

NicolasDorier:
A new tool has been added for convenience `switch-node.sh bitcoinknots`

mp21-org:
Updated the Deployment FAQ accordingly. Please merge if you see fit:
https://github.com/btcpayserver/btcpayserver-doc/pull/1594

PR #991, opened 2025-05-12
Add nolimits

NicolasDorier:


ndeet:
This was always possible with custom fragments anyway. What is the use case to have it as separate fragment?

PR #1003, opened 2025-08-09
Refactor Monero: Use unified Dockerfile, switch to CMD, and update volume path

deverickapollo:


deverickapollo:
@NicolasDorier - I've completed those additional changes to test `btcpayserver/monero:0.18.4.0-2`

Thanks for calling that out!

Issue #938, opened 2024-10-11
nginx access_log enabled


d2ro:
Hi, we're running the latest btcpayserver-docker (with `"log-driver": "journald"` in `/etc/docker/daemon.json`, if that matters). It seems that nginx `access_log` is enabled by default:

```
$ journalctl -e
[...]
Oct 11 12:39:10 example.com f33d0b06b6c1[518]: example.com 172.18.0.1 - - [11/Oct/2024:10:39:10 +0000] "GET / HTTP/2.0" 302 0 "-" "redacted"
Oct 11 12:39:11 example.com f33d0b06b6c1[518]: example.com 172.18.0.1 - - [11/Oct/2024:10:39:11 +0000] "GET /login?ReturnUrl=%2F HTTP/2.0" 200 15834 "-" "redacted"
[...]
```

The logs don't contain external IP addresses, but they do contain User-Agent headers, and they are flooding our logs. Is there a proper way to disable the logging?

Thank you very much in advance!

d2ro:
Thank you very much for merging the fix!

PR #1034, opened 2025-12-21
Disable nginx access_log

d2ro:
This PR removes the `access_log` directives in the server blocks of `nginx.tmpl`, so that the root level `access_log off;` setting is effective for all server blocks.

This is a follow-up of issue #938. Our logs were full of access log lines. These lines have no real use, as they always contain the internal IP address `172.18.0.1`. So I think it is safe to remove them.

### DIFF
--- a/Production/nginx.tmpl
+++ b/Production/nginx.tmpl
@@ -377,7 +377,6 @@ server {
 	{{ if $enable_ipv6 }}
 	listen [::]:80;
 	{{ end }}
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Given to recent hacks related to btcpay, I decided to reactivate the logs in https://github.com/btcpayserver/btcpayserver-docker/commit/d7cbe1ffba99894aed30d0069aca80a96a0daf56

PR #1037, opened 2026-01-13
Add BTCPAY_UPDATE_CLEAN environment option

rockstardev:
This commit provides a less-intrusive pruning of all images after running an update.

Ideally, there would be a better prune filter but that appears to require more granular tracking of images; including btcpayserver/docker-compose-generator (which only pulls from latest).

This option should suffice for the time being.

Original PR by AaronFiore: https://gitea.com/AaronFiore/btcpayserver-docker/commit/bf518660c70445c1a2e66de41583ee4f6c7bc4ab

### DIFF
--- a/btcpay-update.sh
+++ b/btcpay-update.sh
@@ -80,4 +80,6 @@ btcpay_update_docker_env
 btcpay_up
 
 set +e
-[ "$BTCPAY_UPDATE_CLEAN" == true ] && ./btcpay-clean.sh || exit 0
+if [ "$BTCPAY_UPDATE_CLEAN" == true ]; then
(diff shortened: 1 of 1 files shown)


btcpay-update.sh
```diff
@@ -80,4 +80,4 @@ btcpay_update_docker_env
 btcpay_up
 
 set +e
-docker image prune -af --filter "label!=org.btcpayserver.image=docker-compose-generator"
+[ "$BTCPAY_UPDATE_CLEAN" == true ] && ./btcpay-clean.sh || exit 0
```
NicolasDorier: why `exit 0` ? the script finish after anyway

PR #1064, opened 2026-07-14
Add GNU Taler merchant support (opt-add-taler-merchant)

rachyandco:
Adds an opt-add-taler-merchant fragment running the GNU Taler merchant backend used by the BTCPay Taler plugin, with its own postgres.

Public merchant endpoints are served under /taler-merchant/ on the BTCPay host (nginx location keyed on the service name, like bitcoin_rtl; equivalent traefik labels with a stripprefix middleware). Private/management/webui APIs are blocked at the proxy and remain Bearer-token-protected regardless.


NicolasDorier:
@rachyandco can you use the existing postgres instead of popping up a new one?

Also can you document it in README?

rachyandco:
@NicolasDorier both changes are implemented

NicolasDorier:
@rachyandco I mean I want you to update the doc on this part https://github.com/btcpayserver/btcpayserver-docker/blob/master/README.md#generated-docker-compose

rachyandco:
some more cleanup and version pinning

PR #1077, opened 2026-08-13
Use btcpayserver-internal:2.4.3-rc4 image

rockstardev:
## Summary

Point the deployment at the internally built 2.4.3-rc4 image instead of the last public release:

- `docker-compose-generator/docker-fragments/btcpayserver.yml`: default image becomes `btcpayserver/btcpayserver-internal:2.4.3-rc4` (still overridable via `BTCPAY_IMAGE`)
- `contrib/build-all-images.sh`: unchanged build (still builds the public 2.4.2), with an added comment noting that the active image is `btcpayserver/btcpayserver-internal:2.4.3-rc4` from an internal repository that cannot be cloned there
- `README.md`: version table updated accordingly

The image was built and pushed to Docker Hub from the signed `v2.4.3-rc4` tag via GitHub Actions (multi-arch: amd64, arm64, arm/v7).

Note: this is a release candidate published to the internal Docker Hub repository — intended for testing the 2.4.3 RC, not as the final public bump.

#### Test plan

- [x] `docker compose` generated config pulls `btcpayserver/btcpayserver-internal:2.4.3-rc4`
- [x] Instance boots and reports version 2.4.3


Kukks (approved):
ACK

PR #1079, opened 2026-08-20
Bumping RTL to v0.15.11

saubyk:
Bumps Ride The Lightning from `v0.15.10` to `v0.15.11` — the current release, published 2026-08-18.

Release notes: https://github.com/Ride-The-Lightning/RTL/releases/tag/v0.15.11

Changed in the same places as #1071:

- `docker-compose-generator/docker-fragments/bitcoin-lnd.yml`, `bitcoin-clightning.yml`, `bitcoin-eclair.yml` — image tag
- `README.md` — image table row
- `contrib/build-all-images.sh` — the three RTL build blocks

`shahanafarooqui/rtl:v0.15.11` is published multi-arch (amd64, arm32v7, arm64v8), and `https://raw.githubusercontent.com/Ride-The-Lightning/RTL/v0.15.11/Dockerfile` resolves, so the README links stay valid.

Additionally, and unlike #1071, this also updates the three RTL blocks in `contrib/DockerFileBuildHelper/build-all.sh` (second commit). It isn't an output of `update-repo.sh` — that writes only `contrib/build-all-images.sh` and the root `README.md` — but recent bumps (LND 0.21.1 / 0.21.2, LIT) have kept it current by hand, and its RTL blocks had been left behind at `v0.15.4`. They are now byte-identical to the corresponding blocks in `contrib/build-all-images.sh`. Its Core Lightning blocks are also stale (`v25.05` vs `v26.06.1-1` in the fragments); I've left those alone as out of scope here.

Happy to drop the second commit if you'd rather that file only move when the generator is re-run.


NicolasDorier:
just bumped it directly, github was broken.

PR #1080, opened 2026-08-26
Pass offline option to CLN.

ndeet:
There is no new binary out yet and no official confirmation.

Source: https://stacker.news/items/1555439

### DIFF
--- a/docker-compose-generator/docker-fragments/bitcoin-clightning.yml
+++ b/docker-compose-generator/docker-fragments/bitcoin-clightning.yml
@@ -19,6 +19,7 @@ services:
         clnrest-protocol=http
         clnrest-host=0.0.0.0
         grpc-port=9736
+        offline
     volumes:
       - "clightning_bitcoin_datadir:/root/.lightning"
       - "bitcoin_datadir:/etc/bitcoin"


rockstardev:
Added directly on master in https://github.com/btcpayserver/btcpayserver-docker/commit/f3ad9d0e406bc422788910064dbf0fa246ff26fa

jooray:
Heads-up: the variant that landed on master ([f3ad9d0](https://github.com/btcpayserver/btcpayserver-docker/commit/f3ad9d0e406bc422788910064dbf0fa246ff26fa)) writes `offline=1`, and CLN rejects a value on a flag option — `lightningd` exits instead of starting, so the container ends up in a restart loop. The plain `offline` in this PR is the one that works.

Measured against the pinned image, `btcpayserver/lightning:v26.06.1-1`:

```
$ docker run --rm --network none --entrypoint sh btcpayserver/lightning:v26.06.1-1 -c '
    d=$(mktemp -d)
    printf "network=bitcoin\nbitcoin-rpcconnect=127.0.0.1\noffline=1\n" > $d/config
    lightningd --lightning-dir=$d; echo "EXIT=$?"'

lightningd: Config file /tmp/tmp.D7vhVfY6VP/config line 3: offline=1: doesn't allow an argument
EXIT=1
```

Same run with plain `offline`:

```
INFO    lightningd: Started in offline mode!
```

`tools/docker-entrypoint.sh` writes `LIGHTNINGD_OPT` into `$LIGHTNINGD_DATA/config` verbatim, so the fragment value goes straight into the config file.

Worth fixing quickly, because the failure mode is the inverse of the intent: a node that won't start also isn't following the chain, so it can't react to a peer force-closing or publishing a revoked commitment while it's down. `offline` keeps `lightningd` running and watching, with no listening and no reconnects — strictly better than stopped.

Unrelated to this PR but adjacent to [823fd7c](https://github.com/btcpayserver/btcpayserver-docker/commit/823fd7c): dropping the nginx `/clightning-rest/btc/` route closes one way in, but `bitcoin-clightning.yml` still has `ports: - "3010"`, which publishes clnrest on a Docker-assigned host port on `0.0.0.0`. Operators who want clnrest unreachable during the embargo need to block that separately — and note a rule in `DOCKER-USER` matches the *container* port (3010), since packets are already DNAT'd by the time they get there.

PR #1082, opened 2026-08-28
Fix ThunderHub SSO cookie ownership on shared volume

### LINKED ISSUE #1072: ThunderHub SSO login fails with EACCES when .cookie is owned by root
## Summary

The `bitcoin_thub` container runs as `uid=1000 (node)`, while `btcpayserver` runs
as `root`. Both mount the shared volume `lnd_bitcoin_thub_datadir`.

ThunderHub calls `refreshCookie()` on every SSO login and needs **write** access
to `/data/.cookie`. When that file ends up owned by `root:root` (mode `644`),
ThunderHub can read it but not rewrite it, and login fails.

## Symptom

Clicking the ThunderHub link in BTCPay (`Server Settings → Services`) opens a new
tab that immediately closes. No error is shown in the UI — the failure is only
visible in the container log.

## Error

```
Error: EACCES: permission denied, open '/data/.cookie'
    at FilesService.refreshCookie (/app/dist/modules/files/files.service.js:337:19)
    at AuthResolver.getAuthToken (/app/dist/modules/api/auth/auth.resolver.js:181:31)
    at /app/node_modules/@nestjs/core/helpers/external-context-creator.js:67:33
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
```

## Why it is easy to miss

Whether this happens depends on **which container creates the cookie file first**:

- if ThunderHub creates it → owner is `node`, everything works, possibly for years;
- if BTCPay creates it (container recreation, restore from backup, changed start
  order) → owner is `root`, and SSO silently breaks.

On the affected instance the file was dated several months before the problem was
noticed, so the breakage went unnoticed for a long time — the ThunderHub link is
not used daily, and the closing tab looks like a browser quirk rather than an error.

## Reproduce

1. Ensure `/data/.cookie` in the `lnd_bitcoin_thub_datadir` volume is owned by
   `root:root` with mode `644` (e.g. let BTCPay create it, or `chown root:root` it).
2. Open `Server Settings → Services → ThunderHub` and follow the link.
3. The tab opens and closes; `docker logs generated_bitcoin_thub_1` shows the
   `EACCES` trace above.

## Environment

- ThunderHub image: `apotdevin/thunderhub:base-0.15.4`
- Fragment: `opt-add-thunderhub`
- `bitcoin_thub` runs as `uid=1000 (node)`, `btcpayserver` runs as `uid=0 (root)`
- Shared volume: `lnd_bitcoin_thub_datadir` → `/data` (thub) and
  `/etc/lnd_bitcoin_thub_datadir` (btcpayserver)

## Workaround

```bash
chown -R 1000:1000 /var/lib/docker/volumes/generated_lnd_bitcoin_thub_datadir/_data
docker restart generated_bitcoin_thub_1
```

## Suggested fix

Make ownership deterministic instead of dependent on start order — either

- `chown` the thub datadir to `1000:1000` in the fragment / entrypoint, or
- have BTCPay create the cookie file with ownership matching the ThunderHub
  container, or
- create the file with mode `666` (least preferable).

Since the ThunderHub image deliberately drops privileges, aligning the volume
ownership on the BTCPay side seems the cleanest option.

apotdevin:
## Problem

ThunderHub (`bitcoin_thub`) runs as uid 1000 (`node`) while `btcpayserver` runs as root, and both mount `lnd_bitcoin_thub_datadir`. Depending on startup order, the volume directory or the SSO `.cookie` file can end up owned by `root:root`. ThunderHub then fails to rotate the cookie during SSO login (`EACCES: permission denied, open '/data/.cookie'`), and the login tab silently closes.

Fixes #1072

## Fix

Adds a one-shot `bitcoin_thub_dircheck` busybox container (pinned by digest, multi-arch) that chowns the volume directory and the cookie file to uid 1000 before `bitcoin_thub` starts. ThunderHub itself keeps running unprivileged.

### DIFF
--- a/docker-compose-generator/docker-fragments/opt-add-thunderhub.yml
+++ b/docker-compose-generator/docker-fragments/opt-add-thunderhub.yml
@@ -4,11 +4,25 @@ services:
       BTCPAY_BTCEXTERNALTHUNDERHUB: "server=/thub/sso;cookiefile=/etc/lnd_bitcoin_thub_datadir/.cookie"
     volumes:
       - "lnd_bitcoin_thub_datadir:/etc/lnd_bitcoin_thub_datadir"
+  # ThunderHub runs as uid 1000 (node) while btcpayserver runs as root and
+  # shares lnd_bitcoin_thub_datadir. If the volume or the SSO cookie file ends
+  # up owned by root, ThunderHub cannot rotate the cookie and SSO login fails
+  # (https://github.com/btcpayserver/btcpayserver-docker/issues/1072).
+  # This one-shot container fixes ownership before ThunderHub starts.
+  bitcoin_thub_dircheck:
+    image: busybox:1.36.1@sha256:73aaf090f3d85aa34ee199857f03fa3a95c8ede2ffd4cc2cdb5b94e566b11662
+    container_name: generated_bitcoin_thub_dircheck
+    restart: "no"
+    command: ["sh", "-c", "chown 1000:1000 /data && { [ ! -e /data/.cookie ] || chown 1000:1000 /data/.cookie; }"]
+    volumes:
+      - "lnd_bitcoin_thub_datadir:/data"
   bitcoin_thub:
     image: apotdevin/thunderhub:base-0.15.4@sha256:5a36615e4fc300837595156869d67d4c9ef9703d5ead8276911b3e80077b3d3f
     container_name: generated_bitcoin_thub_1
     restart: unless-stopped
     stop_signal: SIGKILL
+    depends_on:
+      - bitcoin_thub_dircheck
     environment:
       NO_VERSION_CHECK: "true"
       COOKIE_PATH: "/data/.cookie"
--- a/docker-compose-generator/docker-fragments/opt-add-thunderhub.yml
+++ b/docker-compose-generator/docker-fragments/opt-add-thunderhub.yml
@@ -17,7 +17,7 @@ services:
     volumes:
       - "lnd_bitcoin_thub_datadir:/data"
   bitcoin_thub:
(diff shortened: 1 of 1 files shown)


BitcoinMitchell:
Thank you for tackling this issue! I've ran into this problems a few time (and resolved it manually a few times).

---

If it's not too rude to ask, any plans on updating the pinned image to a newer ThunderHub version? I tried to update 0.15.4 to 0.18.4 (to try and solve the issue you are currently solving), but ran into some issues: https://github.com/btcpayserver/btcpayserver-docker/pull/1060

It seemed that Thunderhub was ignoring the `BASE_PATH` prefix that is defined in the `thunderhub:base` images. Opening `/thub/sso?token=EXAMPLE5f17fb5d67558f48e81544TOKEN` loaded:
- `/thubapi/config`
- `/graphql`

Which seemed to be missing the `/thub` prefix (and return 200 if you add the prefix). 

I tried using version `apotdevin/thunderhub:base-0.18.4@sha256:5a36615e4fc300837595156869d67d4c9ef9703d5ead8276911b3e80077b3d3f`

apotdevin:
@BitcoinMitchell thanks for bringing this up. I'll fix and bump the version in this same PR (if it hasn't been merged in by then)

Update:
Fixed with commit https://github.com/apotdevin/thunderhub/commit/bf6099131e976b978cb7be9034dcfe5ef74e3803

BitcoinMitchell:
I've just pulled and manually ran `generated_bitcoin_thub_dircheck` (to be sure), but doesn't seem to fix the SSO issue (for my setup). I've fully removed Thunderhub as well (removed volume and all) to no avail. 

`lnd_bitcoin_thub_datadir` doesn't seem to have `/data/.cookie` to begin with, so not even sure if this a Thunderhub issue or BTCPayServer not placing a cookie where it should've. Then again, I've messed around with this so much, it might be a me problem.

Will try and do more investigating later.

PR #1095, opened 2026-09-03
Diagnose NBXplorer shutdown delay

NicolasDorier:
Temporary CI instrumentation to identify why NBXplorer consumes the full shutdown timeout. This PR will be closed after collecting the diagnostic run.


NicolasDorier:
Diagnostic complete: the managed stack shows only NBitcoin P2P foreground threads remaining after NBXplorer host shutdown; PostgreSQL had no active NBXplorer sessions. The fix belongs in NBXplorer.

PR #1097, opened 2026-09-07
Increase bitcoind file descriptor limit

vbrekher:


NicolasDorier:
why? what did you experience without that

vbrekher:
No failure on my side. Docker 29 dropped the container nofile default to 1024, so I figured it made sense to set it explicitly for bitcoind. If that's not a concern for BTCPay, happy to close this.

NicolasDorier:
thanks for the head up, it could have impacted btcpayserver where 500+ are open. But this is fine for bitcon core which has only 50+

PR #1123, opened 2026-09-16
Serve ACME challenges for non-redirecting hosts

NicolasDorier:
## Why

When a virtual host uses `HTTPS_METHOD=noredirect`, its HTTP server currently sends `/.well-known/acme-challenge/` requests to the application backend. This can prevent HTTP-01 certificate validation.

## Implementation

Serve ACME challenge files from `/usr/share/nginx/html` in the non-redirecting HTTP server. The template avoids declaring the location when the legacy ACME companion already supplies host or default vhost configuration, preventing duplicate Nginx locations.


NicolasDorier:
Closing this because the proposed location is redundant with BTCPay's ACME companion flow.

For initial issuance and `HTTPS_METHOD=noredirect`, the companion writes the ACME challenge location into `/etc/nginx/vhost.d/<hostname>` before issuing the certificate and reloads Nginx. The normal HTTP server already includes that file, so `/.well-known/acme-challenge/` is served from the shared webroot rather than proxied to BTCPay Server.

The existing explicit challenge location in the redirect-only HTTP server is still necessary because that server does not include `vhost.d`; it fixes the renewal case after a certificate exists. #1123 adds the location only when no `vhost.d` file exists, which does not help the normal companion flow and is not a reliable indication that another ACME client is managing the shared webroot.

